PR #32, opened 2018-05-21
Change tests organization and other improvements

anler:
This PR addresses the following:

- When running tests the redux logger middleware is not used to keep the tests output clean
- Removes the `cleanup` task and adds a `clean` task that only removes build-generated files
- Moves unit/integration tests for the app into `test/app`
- Fixes the *watch mode* for tests that was broken
- Improves README.md
- Adds a new task `typecheck` that runs the typescript compiler through the app in order to find type errors
- Removes `test/preprocessor` and `test/runTests` files and substitute them with `test/setup` that configures jest. With these changes is no longer required to write `import "../utils/enzymeConfig"` at the beginning of a test, we can import directly from `app/...` e.g: `import Counter from "app/components/Counter"` and finally we can use directly [enzyme matchers](https://github.com/FormidableLabs/enzyme-matchers) when writing tests
- We can also from now use `import ... from "app/..."` instead of using relative paths
- Fixes a wrong definition for `array-type` lint rule
- Renames `app/main.development.js` ⃗ `app/main.electron.js` since it has nothing to do with the node environment


README.md
```diff
@@ -2,22 +2,78 @@
...
+
+### Build
+
+To build the application run: `yarn build` and then `yarn start` to launch the build.
 
-# build electron application for production
-yarn build
+### Test
+
+We use [Jest](https://facebook.github.io/jest/) for testing. All test commands are a thing wrapper around jest so with any of them you can use [Jest's CLI options](https://facebook.github.io/jest/docs/en/cli.html)
```
aesedepece: `thing` → `thin`

PR #31, opened 2018-05-16
Crypto module

kronolynx:
This module will handle keypairs, wallets, etc.

Issue #26, opened 2018-05-14
Wrap storage with encryption


aesedepece:
Inside LevelDB, we should have both plain stores and encrypted stores.

Encrypted stores shall use symmetric encryption (`AES/CBC/PKCS5Padding`) in a [transparent way](https://en.wikipedia.org/wiki/Transparent_Data_Encryption).

Encrypted stores are unlocked using a user-defined passphrase.

aesedepece:
We already produced AES decrypting code with TypeScript in the past for another project:

```typescript
  export function AESdecrypt(ciphertext: number[], key: any): any {
    // Derive AES key from bytes
    const dig = crypto.SHA256(key)
    const keyBytes = dig.words.slice(0, 4)
    const secretKey = crypto.lib.WordArray.create(keyBytes)
    // Extract initialization vector
    const iv = ba2wa(ciphertext.slice(0, 16))
    const bytes = ba2wa(ciphertext.slice(16))
    // Perform decryption
    const decrypted = crypto.AES.decrypt(bytes.toString(crypto.enc.Base64), secretKey, { iv })
    return decrypted
}
```

[(See full source)](https://github.com/stampery/guard-www/blob/2f848627ce482349a6fbedf88dce60dfa12b3ff2/src/ts/monguard/SCRAM.ts#L120)

Issue #25, opened 2018-05-14
Basic LevelDB support and initialization


aesedepece:
Upon start up, Sheikah should read from LevelDB or initialize it if still empty.

PR #33, opened 2018-05-21
Encrypted storage

aesedepece:
A basic implementation of the encrypted storage (see #24, #25, #26)

The beef is in the `Vault` class in`vault.ts`. It wraps LevelDB (levelup + leveldown) with AES symmetric encryption/decryption.

This PR is not yet to be merged in its current form. It first needs to be rebased because of #32 and potentially #31.


app/lib/storage/vault.ts
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Close the vault, the LevelDB database and the underlying storage, freeing
+   *  up any locks and file descriptors.
+   * @returns {Promise<any>}
+   */
+  public async close(): Promise<any> {
+    this.db.close()
+  }
+
+}
```
anler: Here the same, I'm totally ok with it, although I'm more fan on this alternative implementation:

1. Define the interface I'm going to use in the rest of the program to access the key-value store engine, for example (the name of the interface is irrelevant, is just the one that first came to mind):

```typescript
interface IStore {
  put(key: string, value: any): Promise<boolean>;
  get(key: string): Promise<any>;
  close(): void
}
```

2. Define a concrete implementation for that interface, in our case it would be a leveldb-backed implementation:

```typescript
class Vault implements IStore {
  constructor(private name: string, private db: any, private cryptOptions)
  // implementation of methods
  // ....
}
```
where `db` is already an active connection to leveldb

3. Define a factory function that is responsible of creating an instance of the concrete implementation for `IStore`:

```typescript
function createLevelDbVault(name: string, password: string, path?: string): Promise<Vault> {
  return new Promise(async resolve => {
    // ...
    // This causes the `this.ready` promise to resolve.
    resolve(new LevelDbVault(name, db, cryptOptions))
  })
}
```
The reason to use a factory function instead of putting that logic inside the class constructor is to avoid having a component with too much responsibility, if for some reason the way the db connection is created is enriched, all we need to do is create a new factory function to accommodate the change. 

The advantage of this model is that now, whenever I'm testing a component of the application that needs an `IStore` value, I can pass one more suitable for testing, e.g. one that keeps the data in-memory.

test/app/lib/storage/vault.spec.ts
```diff
@@ -0,0 +1,51 @@
+import {Vault} from "../../../../app/lib/storage/index"
```
anler: In #32 I added the [modulePath](https://github.com/witnet/sheikah/blob/master/package.json#L31) option to jest config to be able to write this line as:
```typescript
import {Vault} from "app/lib/storage/index"
```

test/app/lib/storage/vault.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const vaultName = "test-vault"
+const vaultPassword = "password"
+let globalVault: Vault
```
anler: There's an issue here that wouldn't exist if we use what I proposed in the previous comment, and the issue is that Jest runs the tests in parallel, if we define another test that uses `Vault`, the data stored in it would be shared by the tests and that could cause problems

app/lib/storage/encryption.ts
```diff
@@ -0,0 +1,159 @@
...
+  public getKey(salt: Buffer = this.salt): Buffer {
+    return crypto.pbkdf2Sync(
+      this.password,
+      salt,
+      this.iterations,
+      this.keyBytes,
+      this.digest
+    )
+  }
+}
```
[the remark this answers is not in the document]
aesedepece: I'm just a little afraid that the suggested pattern may be harder to grasp for newcomers and eventual contributors.
Nonetheless, I totally love this approach and at the moment I see no other reason for not adopting app-wide. I'll rewrite it.

app/lib/storage/encryption.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.algorithm = opts.algorithm || "aes-256-cbc"
+    this.digest = opts.digest || "sha256"
+    this.iterations = opts.iterations || 100000
+    this.ivBytes = opts.ivBytes || 16
+    this.keyBytes = opts.keyBytes || 32
+    this.password = opts.password
+    this.saltBytes = opts.saltBytes || 32
+
+    this.iv = opts.iv || crypto.randomBytes(this.ivBytes)
+    this.salt = opts.salt || crypto.randomBytes(this.saltBytes)
```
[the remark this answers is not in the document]
aesedepece: By that point (line 56), `this.ivBytes` and `this.saltBytes` just can't be `null` as they are ORed against default values earlier in the constructor. This code is actually safer than it may seem at first sight. 

However, I understand that your concern is about the constructor being modified in the future in any way that could break the type safety of `this.ivBytes` and `this.saltBytes`, for example if they lose their default values.

app/lib/storage/vault.ts
aesedepece: I like your way of thinking...

Same as before, I'll give a try to your suggested pattern. I find it indeed to be quite idiomatic. Forgive me if I got carried away too much by the object-oriented nature of TypeScript :rofl:

test/app/lib/storage/vault.spec.ts
aesedepece: Regardless of any punctualization on wether it's safe or not in its current form, I'm favorable to refactor Vault in such away that it's dettached from the db connection so it can be tested more easily.

aesedepece:
@mmartinbar and @kronolynx would you like to chime in and share some comments on @anler's review? Eight eyes see more than four! :stuck_out_tongue_winking_eye: 

I think the points brought by @anler on his review need to be at least known by everyone in the team. The scope of the discussion transcends the scope of this particular issue indeed. Amongst other topics, there's this dilemma on how much do functions need to be coupled to the data/state they operate upon. The eternal object-oriented vs. functional paradigm fight is knocking at the door!

test/app/lib/storage/vault.spec.ts
anler: crap, that's a pity, I'll see if there's a way of specifying jest config in a way WebStorm likes it

test/app/lib/storage/vault.spec.ts
anler: After testing a bit I think I found the use case where it fails, executing `yarn test` seems to execute tests sequentially, but if I run `yarn jest --watch` and after the first sequential run of tests I press `a` (run all tests) it seems those tests are run in parallel. If you duplicate `vault.spec.ts` file to `vault2.spec.ts` for example, running all the tests in this scenario

screenflow

app/lib/storage/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+/**
+ * Checks asynchronously if a certain directory exist. If it does not, creates
+ * the directory and all missing parents in the path. This is equivalent to
+ * `mkdir -p` in POSIX.
+ * @param {string} path
+ * @returns {Promise<any>}
+ */
+export const ensurePath = async (path: string): Promise<any> => {
```
mmartinbar: Promise<any> could be Promise<boolean> as resolve logic just returns true

app/lib/storage/encryption.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * Crypto buffers deserializer function.
+ * @param {Buffer} data
+ * @returns {Array<Buffer>}
+ */
+const deserialize = (data: Buffer): Array<Buffer> => {
+  const parts: Array<Buffer> = []
+  const l = data.length
```
mmartinbar: for clarity, the name of the variable 'l' could be changed to something like 'len'

app/lib/storage/encryption.ts
```diff
@@ -0,0 +1,159 @@
...
+  const iv = parts[0]
+  const ciphertext = parts[1]
+  const salt = parts[2]
+  const key = opts.getKey(salt)
+
+  const decipher = crypto.createDecipheriv(opts.algorithm, key, iv)
+  const plaintext = decipher.update(ciphertext)
+  const json = Buffer.concat([plaintext, decipher.final()]).toString()
+
+  return JSON.parse(json)
```
mmartinbar: To make the decrypt function symmetric with the encrypt one, it could just return the `Buffer` and leave the parse logic (`JSON.parse`) to the caller.

app/lib/storage/vault.ts
```diff
@@ -0,0 +1,121 @@
...
+    const hashedKey = Encryption.keyHash(key)
+    // Dehydrate/deflate and serialize the object to convert it into a typed
+    // JSON that can be rehydrated/inflated afterwards without losing types.
+    const serializedValue = Buffer.from(JSON.stringify(hydra.dehydrate(value)))
+    // Encrypt the serialized object.
+    const encryptedValue =
+      Encryption.encrypt(serializedValue, this.cryptOptions)
+    // Put the value into the database and wait for write to complete.
+    await new Promise(resolve => {
+      this.db.put(hashedKey, encryptedValue, resolve)
```
mmartinbar: Couldn't/shouldn't this case be handled as in the `get` function? `put` method can also give errors:
`
+      this.db.put(hashedKey, encryptedValue, (error, value) => {
+        if (error) {
+          reject(error)
+        } else {
+          resolve(true)
+        }
+      })
`

app/lib/storage/vault.ts
```diff
@@ -0,0 +1,121 @@
...
+    })
+
+    return true
+  }
```
mmartinbar: Couldn't/shouldn't this case be handled as in the `get` function? `put` method can also give errors:
```
      this.db.put(hashedKey, encryptedValue, (error, value) => {
        if (error) {
          reject(error)
        } else {
          resolve(true)
        }
      })
```

app/lib/storage/utils.ts
anler: Yes, I thought of the same, actually even `boolean` is too broad since the function either fails or returns `true`, so what we really need is a [unit type](https://en.wikipedia.org/wiki/Unit_type), the closest thing in TypeScript would be `void` so the end result would be:
```typescript
export const ensurePath = async (path: string): Promise<void> => {
  return new Promise((resolve, reject) => {
    mkdirp(path, (error: ErrnoException, made: mkdirp.Made) => {
      if (error) {
        reject(error)
      } else {
        resolve()
      }
    })
  })
} 
```

aesedepece:
+ All the commits in this PR have been refactored to adopt the design patterns we have been discussing.
+ The tests are now using a mocked in-memory storage backend.
+ Encryption is tested separately now as well.
+ Commit 2fd666aee77ceb6f6d6b74495c3861a389ae9ba1 has been added. It upgrades the version of `@types/node` we are using.
+ PR rebased, all tests passing, ready to merge.
+ r? @mmartinbar @anler

app/lib/storage/levelBackend.ts
```diff
@@ -0,0 +1,53 @@
...
+export default class LevelBackend implements IStorageBackend {
+
+  /**
+   * A LevelDB connection reference.
+   */
+  private connection: LevelUp
+
+  constructor(connection: LevelUp) {
+    this.connection = connection
+  }
```
anler: Just a suggestion, these lines are equivalent to:
```typescript
export default class LevelBackend implements IStorageBackend {
  constructor(private connection: LevelUp) {
  }
  ...
}
```

PR #35, opened 2018-05-23
Fixes typecheck task

anler:
Fixes `typecheck` task present in `package.json`
- Now the compiler does not emit output when the task is run
- Added option `baseUrl` to tsconfig.json so non-relative imports in
  test files are found when type-checking

## Usage

Just run `yarn typecheck` or `yarn typecheck --watch`


aesedepece:
This is super nice. Kudos!

PR #39, opened 2018-05-25
Improve EncryptedStorage tests

aesedepece:
Tests proving the sucess of `Promise<void>` don't really need to use `expect()` at all.
If they resolve and don't throw any exceptions, they're just OK!

### DIFF
--- a/test/app/lib/storage/encryptedStorage.spec.ts
+++ b/test/app/lib/storage/encryptedStorage.spec.ts
@@ -12,18 +12,21 @@ const globalStorage = new EncryptedStorage(mockedBackend, encryptionSettings)
 
 describe("EncryptedStorage", () => {
   it("should put a string", async () => {
-    const result = await globalStorage.put("foo", "bar")
-    expect(result).toBeUndefined()
+    expect(async () => {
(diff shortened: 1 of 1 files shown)


test/app/lib/storage/encryptedStorage.spec.ts
```diff
@@ -12,18 +12,15 @@ const globalStorage = new EncryptedStorage(mockedBackend, encryptionSettings)
 
 describe("EncryptedStorage", () => {
   it("should put a string", async () => {
-    const result = await globalStorage.put("foo", "bar")
-    expect(result).toBeUndefined()
+    await globalStorage.put("foo", "bar")
   })
```
anler: I prefer to be explicit and specify what exactly the test case is checking, so imo I would prefer something like this:

```js
it("should put a string", () => {
    expect(async () => await globalStorage.put("foo", "bar")).not.toThrow()
})
```

anler: sorry, I got it wrong, revert my comment 😞

aesedepece: No problem! This is reverted in #46

PR #46, opened 2018-05-28
Fix encryptedStorage tests

aesedepece:
Asynchronous tests returning `Promise<void>` do not need any `expect()` assertions, as Jest natively handles exceptions due to uncatched promise rejection.


aesedepece:
@anler r?

PR #47, opened 2018-05-28
Secp256k1 typings

anler:
Changes:
- Add typings for the [secp256k1](https://github.com/cryptocoinjs/secp256k1-node/blob/master/API.md) library API
- Change `lint` task in `package.json` to mimic the `precommit` task, because the first was succeeding when the second was not

### DIFF
--- a/typings/secp256k1/index.d.ts
+++ b/typings/secp256k1/index.d.ts
@@ -1,0 +1,159 @@
+declare module "secp256k1" {
+  /** Options for the `sign` function */
+  export interface SignOptions {
+    /** Nonce generator. By default it is rfc6979 */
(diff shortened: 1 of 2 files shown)


typings/secp256k1/index.d.ts
```diff
@@ -0,0 +1,159 @@
+declare module "secp256k1" {
```
aesedepece: Did you take this from somewhere? We need to acknowledge all 3rd party code (e.g.: by including a header comment) and make sure the licenses are compatible.

anler: I wrote them with my bare hands 😛

PR #48, opened 2018-05-28
Remove semicolons from examples

kronolynx:
There's a point that says `Do not use semicolons` but all the examples had semicolons


aesedepece:
@kronolynx r+. Do you want me to merge it for you?

PR #104, opened 2018-06-08
Bech32 addresses derivation

### LINKED ISSUE #20: Bech32 addresses derivation

kronolynx:
  rationale for why I haven't.

this commit closes #20

### DIFF
--- a/app/lib/crypto/address/p2pkh.ts
+++ b/app/lib/crypto/address/p2pkh.ts
@@ -4,7 +4,15 @@ import {sha256} from "../hash"
 import {ChainType} from "../../chain/chainType"
 
 /**
- * Encode address
+ * Witnet chain prefix
+ */
+enum Prefix {
+  "twit",
+  "wit"
+}
+
+/**
+ * Encode pay-to-public-key-hash (P2PKH) address
  * @param {PublicKey} pubKey public key
  * @param chain
  * @returns {string} address
@@ -17,12 +25,4 @@ export const encode = (pubKey: PublicKey, chain: ChainType): string => {
   const hrp = Prefix[chain]
 
   return Bech32.encode(hrp, b32)
-}
-
-/**
- * Witnet chain prefix
- */
-enum Prefix {
-  "twit",
-  "wit"
 }
--- a/app/lib/crypto/address/p2pkh.ts
+++ b/app/lib/crypto/address/p2pkh.ts
@@ -1,15 +1,28 @@
-import {PublicKey} from "../key/publicKey"
 import * as Bech32 from "bech32"
+import {PublicKey} from "../key/publicKey"
 import {sha256} from "../hash"
+import {ChainType} from "../../chain/chainType"
 
 /**
  * Encode address
  * @param {PublicKey} pubKey public key
- * @param {"wit" | "twit"} hrp human readable part
+ * @param chain
  * @returns {string} address
  */
-export const encode = (pubKey: PublicKey, hrp: "wit" | "twit"): string => {
+export const encode = (pubKey: PublicKey, chain: ChainType): string => {
+  if (!(chain in Prefix)) {
+    throw new Error("Unsupported chain type")
+  }
   const b32 = Bech32.toWords(Buffer.concat([Buffer.from([0]), sha256(pubKey.bytes).slice(20)]))
+  const hrp = Prefix[chain]
 
   return Bech32.encode(hrp, b32)
+}
+
+/**
+ * Witnet chain prefix
+ */
+enum Prefix {
+  "twit",
+  "wit"
 }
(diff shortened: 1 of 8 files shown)


app/lib/crypto/address/p2pkh.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as Bech32 from "bech32"
+import {sha256} from "../hash"
+
+/**
+ * Encode address
+ * @param {PublicKey} pubKey public key
+ * @param {"wit" | "twit"} hrp human readable part
+ * @returns {string} address
+ */
+export const encode = (pubKey: PublicKey, hrp: "wit" | "twit"): string => {
```
aesedepece: Using strings for the hrp argument would only make sense if this function accepted any arbitrary value, for example in the case that we wanted to support other blockchains with different prefixes.

The problem is that we will need to hardcode the same string in many different places across our project, and use something like `encode(publicKey, app.chain === "main" ? "wit" : "twit")` pretty much everywhere. This is suboptimal because it introduces a lot of code repetition and diminish flexibility if we want to change the prefixes at some point.

Instead, I'd suggest `encode(publicKey, app.chain)` where `app.chain` has the following enum type:
```ts
enum ChainType {
  "test",
  "main",
}
```
Then the `encode/2` function should internally map the chain type to an hrp:
```ts
enum Prefix {
    "twit",
    "wit"
}

function encode(pubKey: PublicKey, chain: ChainType): string {
    if (!(chain in Prefix)) { 
        throw new Error("Unsupported chain type")
    }

    const hrp = Prefix[chain]

    /* ... */
}

encode(key, ChainType.main)
```

app/lib/crypto/hash.ts
```diff
@@ -18,3 +18,22 @@ const hmacFunction = (hashFunctionName: string, key: Buffer) => {
 export const sha512hmac = (key: Buffer, data: Buffer) => {
   return hmacFunction("sha512", key)(data)
 }
+
+/**
+ * Create sha function for the indicated hash function
```
aesedepece: More expressive comment:
`Wrap crypto.createHash to streamline its worflow without losing parameterization.`

app/lib/crypto/hash.ts
```diff
@@ -18,3 +18,22 @@ const hmacFunction = (hashFunctionName: string, key: Buffer) => {
...
+ * @param {string} hashFunctionName name of a hash function
+ * @param {Buffer} bytes
+ * @returns {Buffer} sha function
+ */
+const shaFunction = (hashFunctionName: string, bytes: Buffer) => {
```
aesedepece: This function name isn't as expressive as desirable, we should always favor actions/verbs over nouns.

What about `applySHA()`?

aesedepece: Or even just `applyHash()` or `hash()`, as it accepts other algorithms out of the SHA family.

test/app/lib/crypto/address.spec.ts
```diff
@@ -0,0 +1,12 @@
+import {getMasterKey} from "./helpers"
+import * as PublicKey from "../../../../app/lib/crypto/key/publicKey"
+import {encode} from "../../../../app/lib/crypto/address/p2pkh"
+
+describe("p2pkh", () => {
+  const masterKey = getMasterKey("000102030405060708090a0b0c0d0e0f")
+
+  it("should encode address", () => {
```
aesedepece: `it("should derive a Bech32-encoded P2PKH address from an extended public key", () => { `

test/app/lib/crypto/helpers.ts
```diff
@@ -0,0 +1,20 @@
+import {ExtendedKey} from "../../../../app/lib/crypto/key/key"
+import {fromEntropy} from "../../../../app/lib/crypto/seed"
+import * as PrivateKey from "../../../../app/lib/crypto/key/privateKey"
+
+/**
+ * Helper for testing
```
aesedepece: I'm a bit reluctant of including test helpers inside the library itself. Yet I'm not completely against it and can't figure out a better option in this case (moving this into the tests folders may require us to create a `helpers` directory that should be ignored by jest).

test/app/lib/crypto/helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param {string} seedEntropy
+ * @param {string} passPhrase
+ * @returns {ExtendedKey<PrivateKey>}
+ */
+export function getMasterKey(seedEntropy: string, passPhrase = "Bitcoin seed"):
```
aesedepece: `passPhrase` should be `Witnet seed` by default. Only the tests use `Bitcoin seed` in order to recycle existing test vectors.

app/lib/crypto/address/p2pkh.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Encode address
+ * @param {PublicKey} pubKey public key
+ * @param chain
+ * @returns {string} address
+ */
+export const encode = (pubKey: PublicKey, chain: ChainType): string => {
+  if (!(chain in Prefix)) {
+    throw new Error("Unsupported chain type")
```
aesedepece: As a side comment, see #122

PR #132, opened 2018-06-13
feat(crypto): support P2PKH Bech32 address decoding

### LINKED ISSUE #105: feat(crypto): support P2PKH Bech32 address decoding

kronolynx:
  rationale for why I haven't.

closes #105

### DIFF
--- a/app/lib/crypto/address/p2pkh.ts
+++ b/app/lib/crypto/address/p2pkh.ts
@@ -8,8 +8,8 @@ import {Errors} from "../errors"
  * Witnet chain prefix
  */
 enum Prefix {
-  twit,
-  wit
+  twit = ChainType.test,
+  wit = ChainType.main
 }
 
 /**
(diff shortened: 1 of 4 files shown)


app/lib/crypto/address/p2pkh.ts
```diff
@@ -22,8 +22,30 @@ export const encode = (pubKey: PublicKey, chain: ChainType): string => {
...
+
+  let chainCode: ChainType
+
+  if (prefix === "twit") {
+    chainCode = ChainType.test
+  } else if (prefix === "wit") {
+    chainCode = ChainType.main
+  } else {
+    throw new Error(Errors.UNSUPPORTED_CHAIN_TYPE)
+  }
```
aesedepece: Making the most of the magic properties of enums, behold this:
```ts
// Unless `ChainTypes[...]` is used somewhere else, this enum can be made constant.
const enum ChainTypes {
  test,
  main
}

// Note that here I'm making the mapping between ChainTypes and Prefixes explicitly, jut in case
// we change the members order in any of the enums at some point, which would remain unnoticed
// by the compiler and would break all logic.
enum Prefixes {
  twit = ChainTypes.test,
  wit = ChainTypes.main
}

// As Prefixes is not a constant enum, it can be easily queried for member pertenence.
if (!(prefix in Prefixes)) {
  throw new Error(Errors.UNSUPPORTED_CHAIN_TYPE)
}

// This is perfectly safe and the compiler recognizes the type as `ChainType`.
const chainType = Prefixes[prefix]
```

app/lib/crypto/address/p2pkh.ts
```diff
@@ -22,8 +22,30 @@ export const encode = (pubKey: PublicKey, chain: ChainType): string => {
...
+/**
+ * Decode the chain type and the hash of the first 20 bytes of the public key
+ * @param {string} address
+ * @returns {[ChainType , Buffer]}
+ */
+export const decode = (address: string): [ChainType, Buffer] => {
+  const {prefix, words} = Bech32.decode(address)
+  const keyHash = Buffer.from(Bech32.fromWords(words))
+
+  let chainCode: ChainType
```
aesedepece: Hey, this is not a `chainCode` :wink: 
`chainType` would be better.

PR #137, opened 2018-06-13
Travis

anler:
# PR Prelude

- [ ] I have read and understood [CODE_OF_CONDUCT][code] document.
- [ ] I have read and understood [CONTRIBUTING][cont] document.
- [ ] I have read and understood [docs/styleguide][style] document.
- [ ] I have included tests for the changes in my PR. If not, I have included a
  rationale for why I haven't.
- [ ] My code is formatted and is accepted by `yarn fmt-verify`.
- [ ] **I understand my PR may be closed if it becomes obvious I didn't

[code]: https://github.com/witnet/sheikah/blob/master/docs/CODE_OF_CONDUCT.md
[cont]: https://github.com/witnet/sheikah/blob/master/docs/CONTRIBUTING.md


package.json
```diff
@@ -21,12 +21,7 @@
     "start-electron": "electron ./dist/main",
     "package": "yarn build && build --publish never",
     "package-dev": "yarn build-dev && build --publish never",
-    "package-win": "yarn build && build --win --x64",
-    "package-win-dev": "yarn build-dev && build --win --x64",
-    "package-linux": "yarn build && build --linux",
-    "package-linux-dev": "yarn build-dev && build --linux",
-    "package-all": "yarn build && build -mwl",
-    "package-all-dev": "yarn build-dev && build -mwl",
+    "travis": "yarn lint && yarn fmt-verify && yarn test-all && yarn package",
```
aesedepece: Note that this is not 1:1 equivalent to what we had before. Travis will now be [fail-fast], so if you have errors in more than one of the steps you will need to do more back an forth to learn about them and get everything right.

However, as now you can run `yarn travis` and replicate the whole Travis flow locally at once, I don't see it as a problem :ok_hand: .

[fail-fast]: https://en.wikipedia.org/wiki/Fail-fast

anler: I added the fail-fast option on purpose to receive earlier feedback from Travis if something with the styling and the like broke, I prefer this because I can always run all the commands in my machine to see what else is borken. What do you think of having two tasks: `travis` which runs the checks in parallel (useful for running it locally in our dev machines) and a new `travis-seq` which is the one used by Travis and that runs the checks with fail-fast?

aesedepece: Sounds good!

PR #140, opened 2018-06-13
build(npm): fix node version to match Electron's

anler:
# PR Prelude

  rationale for why I haven't.

- set `engines` in `package.json` to match the Node version used by Electron
- remove `.node-version` since the check in `package.json` is sufficient.
- leave `engineStrict: false` in `package.json` to allow development with greater Node versions,  but still npm will issue a warning if the version used is different.

[code]: https://github.com/witnet/sheikah/blob/master/docs/CODE_OF_CONDUCT.md
[cont]: https://github.com/witnet/sheikah/blob/master/docs/CONTRIBUTING.md


aesedepece:
Surprisingly, yarn is erroring on Node versions different than `^8.x.x` (in spite of `engineStrict: false`).

We may need to update the build matrix to only test on node 8.

anler:
@aesedepece should we leave only node 8 in the build matrix?

aesedepece:
At the end of the day, the main process is always being run on Node 8 because of Electron, so there's no need to test on other versions apart from checking if we are ready for those when Electron supports them.

Issue #168, opened 2018-06-15
fix(webpack): fix yarn dev app.html not found


mmartinbar:
**Describe the bug**
app.html not found by main process in `dist` directory when executing `yarn dev`from a clean environment.

**To Reproduce**
`yarn clean`
`yarn dev`

**Expected behavior**
app.html should be found by main process and electron app should be seen.

aesedepece:
Closed by #169

PR #169, opened 2018-06-15
fix: fix app.html not found when running yarn dev

anler:
# PR Prelude

  rationale for why I haven't.
- [X] My code is formatted and is accepted by `yarn fmt-verify`.

when running yarn dev the Electron app was complaining that app.html could not be found, the reason
was that yarn dev uses the hot-server which doesn't emit the build files (with the writeToDisk:
true, it only emits the files after stopping the server which is too late) so the solution is to
detect in main/index.js if the app is running in HOT mode in which case it looks for app.html in the
original place rather than in the dist directory

#168 

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -1,8 +1,14 @@
 const { app, BrowserWindow, Menu, shell } = require('electron');
+const path = require('path'); // eslint-disable-line
+const paths = require('../../config/paths')
 
 let menu;
 let template;
 let mainWindow = null;
+
+const appHtml = process.env.HOT ?
+      `file://${paths.templateAppHtml}` :
+      `file://${paths.filenameAppHtml}`;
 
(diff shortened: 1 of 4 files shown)


app/main/index.js
```diff
@@ -1,17 +1,22 @@
...
+const paths = require('../../config/paths')
 
 let menu;
 let template;
 let mainWindow = null;
 
+const appHtml = process.env.HOT ?
+      `file://${paths.templateAppHtml}` :
+      `file://${paths.filenameAppHtml}`;
+
```
aesedepece: This is super cool! :sunglasses:

aesedepece:
Merged in 4c0921a66cd42a09e5937fb84efd5446090756ab

aesedepece:
This closes #168

PR #195, opened 2018-06-28
feat(subsystems): add PlainCipher

### LINKED ISSUE #183: feat(subsystem): add NoCipher

aesedepece:
Cipher implementation for no encryption.

Yes, this is just a dummy pass through that safely works over any type.

All storages must use a `Serializer`, a `Cipher` and a `Persister`. This  will be used by the general app settings persistent storage, which is unencrypted. 

No tests included for obvious reasons :rofl:

fix #183

### DIFF
--- a/app/main/ciphers/plain.ts
+++ b/app/main/ciphers/plain.ts
@@ -1,0 +1,28 @@
+import Cipher from "./index"
+
+/**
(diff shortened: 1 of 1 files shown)


aesedepece:
Merged along #201

Issue #182, opened 2018-06-27
[MEGA-EPIC] Overall prototype app flow


aesedepece:
Note: [r] means renderer process and [m] means main process :wink: 

#### Main process startup routine

1. [m] Start app by running the entry point at `/app/main/index.ts`
2. [m] Start `appSystem`
3. [m] Create persistent storage
4. [m] Reconstruct app-level persistable state from storage
5. [m] Create IPC channels
6. [m] Create main Electron window

#### Renderer process startup routine
1. [r] Configure Redux store
2. [r] Query main process for app-level state and put it into store.
3. [r] Render app root

#### FTU _(first time usage)_
1. [r] Render wallet creation wizard welcome screen, wait input
2. [r] Render wallet type selection, wait input
- [r] If `new`: (see #217)
2.1.1. [r] Tell [m] to create a new wallet
2.1.2. [m] Create new wallet, link it from app state as unconsolidated, return app state updater
2.1.3. [r] Apply app state updater
2.1.4. [r] Render seed backup, wait input
2.1.5. [r] Render seed confirmation/importer, wait input
2.1.6. [r] Send seed to [m]
2.1.7. [m] Validate seed, return updater
2.1.8. [r] Render wallet caption entry, wait input or skip
2.1.9. [r] Render wallet password entry, wait input
2.1.10. [r] Send password and caption (if any) to [m]
2.1.11. [m] Consolidate wallet, persist encrypted wallet, persist state, return updater
2.1.12. [r] Apply app state updater
2.1.13. [r] Tell [m] to generate 100 addresses for m/3'/4919'/0'/0, render spinner
2.1.14. [m] Generate addresses, persist wallet, return updater
2.1.15. [r] Apply app state updater
- [r] If `import`: (see #218)
2.2.1. [r] Go to 2.1.5 through 2.1.14
2.2.2. [r] Ask [m] for related TXs and UTXOs.
2.2.3. [m] Ask `rust-witnet` for related TXs and UTXOs, add to wallet, persist wallet, return udpater
2.2.4. [r] Apply app state updater
- [r] If `master key`: (see #219)
2.3.1. [r] Render xpriv import, wait for input
2.3.2. [r] Send xpriv to [m]
2.3.3. [m] Create new wallet, link it from app state as unconsolidated, return app state updater
2.3.4. [r] Apply app state updater
2.3.5. [r] Go to 2.1.8
- [r] If `hardware device`: unimplemented
[...]

PR #215, opened 2018-07-11
Huge step towards completion of #182

aesedepece:
  rationale for why I haven't.

This PR contains a number of changes geared toward completeness of the overall app workflow as defined in #182:
+ Makes the renderer process query the main process for existing wallets.
+ Initializes the Redux store with the received wallets info. 
+ Exposes the Redux store to routes.
+ Creates the possibility to use conditionally rendered routes (not based on path matching but on more complex logic).
+ Creates subsystems in the main process for holding app info and wallets info.
+ Adds an `appStateManager` that manages access and update of app state.
+ Introduces storage initializers, so that we can initialize local databases with default data.
+ Upgrades Electron to v2.0.4.
+ Eases debugging with WebStorm.
+ Refactors a few things and fixes some others.
+ Removes all boilerplate code.


app/main/api/handlers/getWallets.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Handler function for "getWallets" methods.
+ * It retrieves and returns a list of wallet info objects taken from app state.
+ * @param system
+ * @param params
+ */
+export default async function getWallets(system: SubSystems, params: any) {
+  const _wallets = system.appStateManager.get("wallets")
+
+  return asRuntimeType(_wallets, Wallets, Contexts.STATE)
```
anler: I don't understand, this is the picture I have in mind regarding the decoding/encoding of complex types inside handlers:

```
              +---------------------------------------+
    request   |  +-------------------+                |
  ------------+-->  asRuntimeType    |                |
              |  +-------|-----------+                |
              |  +-------v-----------+                |
              |  | logic that works  |                |
              |  | with complex      |                |
              |  | business type     |                |
              |  +-------|-----------+                |
    response  |  +-------v-----------+                |
  <-----------+--+  asObject         |                |
              |  +-------------------+                |
              |                                       |
              |                          a handler    |
              |                                       |
              +---------------------------------------+
```

for me, `system.appStateManager.get("wallets")` should always return the business type we use to represent wallets, `asRuntimeType` should only be used to decode foreign data which in this case is `params`, and the handler should return `asObject` since that guarantees the response is json serializable

app/common/runtimeTypes/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  IPC = "ipc",
+}
+
+/**
+ * Decode and validate a mixed JS object as a certain runtime type.
+ * @param input
+ * @param runtimeType
+ * @param context
+ */
+export function asRuntimeType<T>(input: t.mixed, runtimeType: t.Type<T>, context?: Contexts) {
```
anler: I would:
- add an explicit return type for this function
- maybe connect the raised error somehow to [InvalidParamsError](https://github.com/witnet/sheikah/blob/3d7b4a59fe6abd0aa2c0f080e1eee0f0d6a99006/app/common/ipc-protocol.ts#L254) so decoding inside a handler automatically returns the appropriate error type

app/main/api/handlers/getState.ts
```diff
@@ -0,0 +1,11 @@
+import { GetStateParams, GetStateResponse } from "app/common/runtimeTypes/ipc/state"
+import { SubSystems } from "app/main/system"
+
+/**
+ * AppInfo handler
+ */
+async function getState(system: SubSystems, params: GetStateParams): Promise<GetStateResponse> {
```
anler: I don't get how could `GetStateParams` work here since params is always an `JsonSerializable` (this actually must be fixed because I got it wrong -I'll open an issue-). There's no way handlers can get their params deserialized because only the handler knows to which business type the params must be deserialized to. So each handler will always receive a `params: JsonSerializable` and will have to deserialize that into the appropriate business type.

This is how each handler should look like:
```
              +---------------------------------------+
    request   |  +-------------------+                |
  ------------+-->  asRuntimeType    |                |
              |  +-------|-----------+                |
              |  +-------v-----------+                |
              |  | logic that works  |                |
              |  | with complex      |                |
              |  | business type     |                |
              |  +-------|-----------+                |
    response  |  +-------v-----------+                |
  <-----------+--+  asObject         |                |
              |  +-------------------+                |
              |                                       |
              |                          a handler    |
              |                                       |
              +---------------------------------------+
```

There's actually an automatic way of doing this by using decorators and reflect-metadata but that would force us to write the handlers as methods in a class because decorators won't work with plain functions.

app/main/index.ts
```diff
@@ -20,6 +18,7 @@ app.on("will-quit", stopApplication)
 if (inDevelopment) {
   const log = require("electron-log")
   log.transports.console.level = log.transports.file.level = "debug"
+  app.commandLine.appendSwitch("remote-debugging-port", "9222")
```
anler: :clap:

app/main/routines/startup.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Initialize app state.
+ * @param system
+ */
+async function stateStart(system: SubSystems) {
+  // Get and parse wallets from storage
+  const _wallets = system.appStorage.get("wallets")
+  const wallets = asRuntimeType(_wallets, Wallets, Contexts.STORAGE)
+  console.log(wallets)
```
anler: We shouldn't use `console.log`, if this is a debug log we would like to have you should use `log.debug` instead (using our `logging` module).

Also, why do we read the wallets and do nothing with them? shouldn't they be added to the app state manager?

app/main/api/handlers/getWallets.ts
aesedepece: Yay, you are right, this assertion shouldn't be needed.

This is a leftover from a time when `AppStateManager._state` was private and had no getter.

As it now has a getter, we can directly do this in a typesafe manner:
```ts
return asObject(system.appStateManager.state.wallets, Wallets)
```

Let me patch it.

app/main/api/handlers/getState.ts
aesedepece: I also thought of decorators when I came with this, but I found the same roadblock as you mentioned: they can only be used on class methods because of weird JS ins and outs.

app/main/routines/startup.ts
aesedepece: Oops... I left that `console.log` behind me, my bad!

As per the other issue, you are right, line `40` should read:
```ts
system.appStateManager.update({ appInfo: config.appInfo, wallets })
```

app/renderer/containers/Root.tsx
```diff
@@ -16,7 +16,7 @@ const Root = ( { store, history }: IRootType) => {
   return (
     <Provider store={store}>
       <ConnectedRouter history={history}>
-        <Routes />
+        <Routes store={store} />
```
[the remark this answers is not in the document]
aesedepece: You are right, I'll move this into our `ui` folder. No containers nor components should exist out of there.

app/main/subsystems/appState.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * App state subsystem.
+ * This simply initializes and returns an app state manager.
+ */
+export class AppStateSubSystem implements Lifecycle<AppStateManager, SubSystems> {
+
+  /**
+   * A reference to the app state manager object itself.
+   */
+  private manager: AppStateManager
```
anler: I got confused by this because I was expecting TypeScript to complain since `manager` is uninitialized, in which case the type should be `AppStateManager | null`. Maybe we should enable also [strictPropertyInitialization](https://blog.mariusschulz.com/2018/05/20/typescript-2-7-strict-property-initialization). Actually enabling this property teaches us something, what we are doing here is pure initialization, which is a bit different than lifecycle, there's no state that needs to start or stop, so this is identical to the [pure json component](https://github.com/witnet/sheikah/blob/master/app/main/subsystems/json.ts#L5)

aesedepece: Now that you mention it... would to the following `pureSubsystem` be 100% equivalent to an instance of my verbose `AppStateSubSystem` class?
```ts
const appStateSubSystem = pureSubsystem(() => new AppStateManager())
```
If so, I support going the `pureSubsystem` way to reduce code repetition.

app/common/runtimeTypes/ipc/state.ts
```diff
@@ -0,0 +1,8 @@
+import { AppState } from "app/main/appState"
```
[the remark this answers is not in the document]
aesedepece: Ok, we can refactor all of this after merging. Feel free to post an issue and take it :stuck_out_tongue_winking_eye:

app/main/subsystems/appState.ts
anler: yes it is hahaha, the idea of the lifecycle is for components where initialization _is not enough_ (connecting to a db, reading some config remotely, etc), components that only need initialization is what I call "pure component or subsystem" :grin:

PR #251, opened 2018-07-18
feat(store): expose ipcRenderer API client to Redux

### LINKED ISSUE #239: feat(store): expose ipcRenderer API client to Redux actions
We need to expose an already instantiated ipcRenderer API client to our Redux actions.

This should easily be achieved thanks to the `thunk` middleware:

Add in `composeEnhancers`:
```ts
applyMiddleware(thunk.withExtraArgument(services))
```

Where services is a sort of renderer process' rough equivalent to the `SubSystems` objects we have in the main process:

```ts
type Services = {
  client: Client
 // whatever else we want redux to have access to
}
```

Then, our action creator functions / thunks can be written as:
```ts
const getWalletsActionCreator = () => async (dispatch, getState, services) => {
  api.getWallets(services.client)
    .then(/* dispatch successful action from here */)
    .catch(/* dispatch errored action from here */)
}
```

This nice functionality of `redux-thunk` is documented in [their README][thunk].

[thunk]: https://github.com/reduxjs/redux-thunk#injecting-a-custom-argument

aesedepece:
  rationale for why I haven't.
- [x] **I understand my PR may be closed if it becomes obvious I didn't actually perform all of these steps.**

This leverages `redux-thunk` extensibility to inject a "services" object into action creator functions. This object only provides apiClient at the moment, but it can hold whatever other objects we may need to use in Redux action creators in the future.

More info about the solution can be found in #239.

fix #239

### DIFF
--- a/app/renderer/store/configureStore.production.ts
+++ b/app/renderer/store/configureStore.production.ts
@@ -1,9 +1,10 @@
-import { createStore, applyMiddleware, compose } from "redux"
-import thunk from "redux-thunk"
-import { createHashHistory } from "history"
-import { connectRouter, routerMiddleware } from "connected-react-router"
-import { createLogger } from "redux-logger"
 import rootReducer, { StoreState } from "app/renderer/reducers"
+import { Services } from "app/renderer/services"
+import createHashHistory from "history/createHashHistory"
+import { applyMiddleware, compose, createStore } from "redux"
+import { createLogger } from "redux-logger"
+import { connectRouter, routerMiddleware } from "connected-react-router"
+import thunk from "redux-thunk"
 
 const logger = (createLogger as any)({
   level: "info",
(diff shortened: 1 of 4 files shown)


app/renderer/store/configureStore.production.ts
```diff
@@ -13,25 +14,25 @@ const logger = (createLogger as any)({
...
+ * @param services
+ */
+function configureStore(initialState: StoreState, services: Services) {
 
-/* eslint-enable no-underscore-dangle */
-const enhancer = compose(
-  applyMiddleware.apply(undefined, middlewares)
-)
+  const middlewares = [thunk.withExtraArgument(services), router]
+  if (process.env.NODE_ENV === "development") {
```
mmartinbar: Development branches are unreachable in production file.

aesedepece:
I left the `middlewares` array in place just for the sake of uniformity of the `configureStore` function across environments and to allow future conditional middlewares in production too.

PR #276, opened 2018-07-26
feat(wallet): set encryption password for new wallet

### LINKED ISSUE #223: feat: set encryption password for new wallet
The renderer process needs a simple function with this contract:
```ts
type EncryptWalletParams = {
  id: string
  password: string
  caption?: string
}

type EncryptWalletResponse = void

encryptWallet(id: string, password: string, label?: string): Promise<EncryptWalletResponse>
```

This function makes the API client issue a request with method "`encryptWallet`" and params of type `EncryptWalletParams`.

The `encryptWallet` handler:
+ uses @aesedepece's code to create a new `jsonAesLevel` wallet storage,
+ (wishlist) checks if another storage existed in Level for same wallet id<sup>1</sup>,
+ puts the wallet storage into the `walletStorageCollection` subsystem,
+ if there's a `caption` in the request, appends it to the unconsolidated wallet entry,
+ write the data of the unconsolidated wallet entry into the wallet storage<sup>2</sup>,
+ consolidates the unconsolidated wallet entry by moving it into the main process' app state object, in the `wallets` array,
+ saves the main process' persistable app state<sup>3</sup>,
+ returns a `WalletInfo` object.

(<sup>1</sup>) _this implies giving the user the chance to choose through a prompt if they want to import the old wallet or overwrite it with the new one. Both will be likely identical because IDs are hashes, so the real question is whether to destroy metadata (stuff that unlike used addresses or the TX history, can't be recovered from the chain)_
(<sup>2</sup>) _indeed, there's nothing too interesting to write to storage at this time, other than the mnemonics._
(<sup>3</sup>) _the main process' app state may contain some data that can be persisted to disk and some data that should not. Every time we want to persist it, it needs first to go through a reducer function that "prunes" all runtime data and leaves only the persistable items._

kronolynx:
  rationale for why I haven't.
- [x] My code is formatted and is accepted by `yarn fmt`.

closes #223


app/common/runtimeTypes/ipc/wallets.ts
```diff
@@ -71,3 +71,45 @@ export type GetWalletError = t.TypeOf<typeof GetWalletError>
...
+  kind: t.literal("SUCCESS"),
+  wallet: Wallet
+}, "EncryptWalletSuccess")
+
+export const encryptWalletErrors = {
+  STORAGE_CREATION_FAILURE: t.literal("STORAGE_CREATION_FAILURE"),
+  WRONG_TYPE_PARAMS: t.literal("WRONG_METHOD_PARAMS"),
+  UNCONSONSOLIDATEDWALLET_UNAVAILABLE: t.literal("UNCONSONSOLIDATEDWALLET_UNAVAILABLE"),
+  INVALID_WALLET_ID: t.literal("INVALID_WALLET_ID"),
+  WALLET_CREATION_FAILURE: t.literal("WALLET_CREATION_FAILURE"),
```
aesedepece: This error is not used anywhere

app/common/runtimeTypes/ipc/wallets.ts
```diff
@@ -71,3 +71,45 @@ export type GetWalletError = t.TypeOf<typeof GetWalletError>
...
+  ENCODE_FAILURE: t.literal("ENCODE_FAILURE"),
+  WALLET_STORE_FAILURE: t.literal("WALLET_STORE_FAILURE"),
+  WALLET_REPLACE_FAILURE: t.literal("WALLET_REPLACE_FAILURE"),
+  WRONG_UNCONSOLIDATED_WALLET: t.literal("WRONG_UNCONSOLIDATED_WALLET"),
```
aesedepece: This error is not used anywhere

app/main/subsystems/wallets.ts
```diff
@@ -12,7 +12,14 @@ import { Lifecycle } from "app/main/lifecycle"
 export class WalletStorage {
 
   /** The JsonAesLevelStorage */
-  private storage: Storage<Buffer, JsonSerializable, Buffer, Buffer> | undefined = undefined
+  private _storage: Storage<Buffer, JsonSerializable, Buffer, Buffer> | undefined = undefined
```
aesedepece: I'm unsure about usage of underscores for private class properties, as we have started using underscores to tell the compiler not to verify unused variables.

Maybe we could name them as `whateverReference` (`storageReference` in this case) if they are references and `whateverValue` if they are values.

app/main/utils/utils.ts
```diff
@@ -33,6 +36,19 @@ export function inject<T, U, V>(fn: (arg: T, injected: V) => U, injected: V) {
   return (arg: T): U => fn(arg, injected)
 }
 
+/**
+ * Wrapper of asRuntimeType that replaces the thrown error with a custom error
+ * @param input
+ * @param runtimeType
+ * @param error error to throw in case of failure
+ */
+export function asType<T, U, V>(input: t.mixed, runtimeType: t.Type<T, U>, error: V): T {
```
aesedepece: Awesome :100:

PR #312, opened 2018-08-07
feat(handlers): make newWallet handler accept xprv seeds

### LINKED ISSUE #260: feat(handlers): make newWallet handler accept xprv seeds
The `newWallet` handler should expect `xpriv` if `mnemonics` isn't provided:
```ts
type NewWalletParams = { mnemonics: string } | { xprv: string }
```

This implies changes to the `unconsolidatedWallet` structure requiring either `mnemonics` or `xpriv` to be defined.

This also requires modifying the `Wallet` shape to make `mnemonics` optionals, as they can't be reconstructed from the `xpriv`.

kronolynx:
  rationale for why I haven't.
- [x] My code is formatted and is accepted by `yarn fmt`.

closes #260

### DIFF
--- a/app/common/config/common.ts
+++ b/app/common/config/common.ts
@@ -8,7 +8,7 @@ export const commonConfig: Partial<Config> = {
     env: inDevelopment ? "dev" : "prod"
   },
   mnemonicsIdGeneration: {
-    salt: "sheikah mnemonics",
+    salt: "sheikah seed",
     hashIterations: 4096,
     keyByteLength: 32,
     hashFunctionName: "sha256"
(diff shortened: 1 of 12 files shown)


app/main/api/handlers/importSeed.ts
```diff
@@ -0,0 +1,218 @@
...
+    .then(seedFromMnemonics)
+}
+
+/**
+ * Generates an ID and builds and returns an UnconsolidatedWallet
+ * @param seed
+ */
+async function buildUnconsolidatedWallet(seed: Seed) {
+  return {
+    id: generateId(seed),
```
aesedepece: We should not generate the Id here, that's `encryptWallet's` business now! :wink:

app/common/config/common.ts
```diff
@@ -8,7 +8,7 @@ export const commonConfig: Partial<Config> = {
     env: inDevelopment ? "dev" : "prod"
   },
   mnemonicsIdGeneration: {
```
aesedepece: This would be better named `walletIdGeneration`.

PR #325, opened 2018-08-09
fix(handlers): properly handle encryptWallet errors

### LINKED ISSUE #324: encryptWallet silently fails when an invalid password is used
If an invalid password is used the front end times out

the next error is not properly handled in encryptWallet
```
InvalidParamsError {
  name: 'InvalidParams',
  message: 'Got a non-compliant JsonAesLevelStorageParams: {"password":"a"}' 
}
```

kronolynx:
- [ ] I have read and understood [CODE_OF_CONDUCT][code] document.
- [ ] I have read and understood [CONTRIBUTING][cont] document.
- [ ] I have read and understood [STYLEGUIDE][style] document.
- [ ] I have included tests for the changes in my PR. If not, I have included a
  rationale for why I haven't.
- [ ] My code is formatted and is accepted by `yarn fmt`.
- [ ] **I understand my PR may be closed if it becomes obvious I didn't

closes #324


app/main/api/handlers/encryptWallet.ts
```diff
@@ -224,7 +221,7 @@ function createKeyChain(keyPath: string): KeyChain {
  * @param keyPath
  */
 function asKeyPath(keyPath: string): KeyPath {
-  return asRuntimeType(keyPath, KeyPath)
+  return asType(keyPath, KeyPath, encryptWalletErrors.INVALID_KEY_PATH)
```
mmartinbar: Love this wrapper, so useful :+1:

app/main/api/handlers/encryptWallet.ts
```diff
@@ -38,13 +37,9 @@ export default async function encryptWallet(system: AppStateS & WalletStorageS,
...
-    .then(async (p) => Promise.all([
-      Promise.resolve(p)
-        .then(inject(newWalletStorage, system.storageFactory)),
-      Promise.resolve(p)
-        .then(inject(getUnconsolidatedData, system.appStateManager))
-        .then(inject(newWallet, system.appStateManager))
-    ]))
+    .then(inject(getUnconsolidatedData, system.appStateManager))
+    .then(newWallet)
+    .then(inject(newWalletStorage, system.storageFactory))
```
mmartinbar: I think this logic misses the part where the plain storage is updated with the WalletInfo (id and caption) of the newly created wallet. I believe this is the point where it should be done because it is not until this point that the id is generated, we don't have it before.

The behaviour shown is that the encrypted wallet is stored in a new folder in the filesystem but when restarting Sheikah app, no wallet is detected because the plain storage is empty.

I'll create a separate issue to refer to this.

mmartinbar: I've created issue #349 to talk about this. I think we should talk to the rest of the team to raise its priority if everything goes right with this PR and the other one we have pending.

app/main/api/handlers/encryptWallet.ts
```diff
@@ -38,13 +37,9 @@ export default async function encryptWallet(system: AppStateS & WalletStorageS,
...
     .then(storeWallet)
     .then(inject(replaceWallet, system))
```
mmartinbar: In #349 I've also mentioned some issues regarding the error handling in the promise chain. I think it is quite important that in the case of error at some point in this chain, we revert whichever actions need to be reverted. For instance, if there is an error in this line `.then(inject(replaceWallet, system))`, then we need to revert the actions performed at the previous steps. For the time being, I think that the only step that would need some revert action would be the storeWallet. However, when we add the storage of the WalletInfo in the plain storage, more 'undoing' actions will be required.

PR #359, opened 2018-08-14
feat(prefilled wallet): add prefilled wallet data

### LINKED ISSUE #330: Define pre-filled wallet initial dataset
#### Motivation
As mandated by Epic #309, we need to define an initial dataset for wallets being pre-filled.

#### Methodology
_Someone tackling this issue shall:_
- Learn about the shape of the `Wallet` type. @kronolynx can provide help understanding it.
- Find examples, fixtures and hard-coded examples of wallets in the current code base.
- Compare those examples with the data depicted in the sketches (see issue #16)
- Augment the examples with more relevant data ranging different states for transactions, addresses etc. so that we can showcase the UI at its best.

#### Deliverables
_A potential PR closing this issue shall provide:_
- A `Wallet` object constructed as either a constant or as a JSON we can import.

anler:
# PR Prelude

  rationale for why I haven't.
- [X] My code is formatted and is accepted by `yarn fmt`.

Add pre-filled wallet data. The data is defined inside `app/renderer/api/getWallet.ts` file and the idea is that the data is only used by the `getWallet` function, no one else should depend on that data (so we can remove easily in the future). Also I copied the mechanism used to generate wallet ids in the tests to here because production code shouldn't depend on test code (so changing the later cannot break the former)

Please, comment any fix, enhancement we should add to the defined data.


app/renderer/prefilledWallet.ts
```diff
@@ -0,0 +1,84 @@
...
+      finalKeys: [finalKey],
+    },
+    {
+      keyPath: [...keyPath, 1],
+      finalKeys: [finalKey],
+    },
+    {
+      keyPath: [...keyPath, 2],
+      finalKeys: [finalKey],
+    }
```
mariocao: Perhaps, it would be better to call `generateFinalKey(keyPath, index)` with the proper index value.

app/renderer/prefilledWallet.ts
```diff
@@ -0,0 +1,84 @@
...
+  const masterSecret = Buffer.from(
+    [196, 188, 49, 56, 180, 80, 166, 118, 177, 247, 66, 74, 173, 113, 48, 127, 194, 4, 164, 0,
+      70, 14, 21, 211, 113, 209, 238, 121, 124, 251, 139, 168])
+
+  const id = crypto.pbkdf2Sync(
+    Buffer.concat([chainCode, masterSecret]),
+    "sheikah seed", 4096, 32, "sha256"
+  ).toString("hex")
+  const _v = CURRENT_WALLET_VERSION
+  const caption = "My prefilled wallet"
```
mmartinbar: Please, use `prefilledWalletCaption` here

app/renderer/prefilledWallet.ts
```diff
@@ -0,0 +1,84 @@
...
+  const epochs = {}
+
+  const seed = { seed: { chainCode, masterSecret }, kind: ("Wip3" as "Wip3") }
```
mmartinbar: I would rename this variable as `seedInfo` to make it clear that it is not just a seed (master secret and chain code)

aesedepece: 100% truth. As defined in [wallets.ts](https://github.com/witnet/sheikah/blob/master/app/common/runtimeTypes/storage/wallets.ts), this is totally a `SeedInfo` and can/should even be statically typed as such.

mmartinbar:
Please @anler, close this PR as it has already been merged in #368 and #372 by @kronolynx, thanks!

PR #412, opened 2018-08-21
fix(issue 364): display error messages in login

### LINKED ISSUE #364: fix(login-form): strange behaviour when unlocking a wallet
**Describe the bug**
Something strange happens sometimes when trying to unlock a wallet several times with a wrong password. The first time the error that appears is "INVALID_WALLET_TYPE" (which is correct) but the second time (or sometimes after a few times), what happens is that the request takes a bit longer and in the end it fails due to a "GENERIC_IPC_ERROR" 

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a wallet with Sheikah
2. Restart Sheikah so that the login form appears and that wallet can be selected
3. Introduce a wrong password for the wallet. See that the "INVALID_WALLET_TYPE" error appears
4. Retry the unlocking of the wallet (just by clicking again the "Unlock" button, without moving away from the form or introducing a new invalid password). See that the spinner shows and that the "GENERIC_IPC_ERROR" error finally appears

anler:
# PR Prelude

  rationale for why I haven't.
- [X] My code is formatted and is accepted by `yarn fmt`.

- add a new `replaceWith` method to the storage that will close the current storage before creating a new one
- show error messages instead of the error type in the login form

Fixes #364

### DIFF
--- a/app/common/runtimeTypes/ipc/wallets.ts
+++ b/app/common/runtimeTypes/ipc/wallets.ts
@@ -65,9 +65,19 @@ export type GetWalletSuccess = t.TypeOf<typeof GetWalletSuccess>
 export const getWalletErrors = {
   INVALID_PARAMS_TYPE: t.literal("INVALID_PARAMS_TYPE"),
   INVALID_WALLET_TYPE: t.literal("INVALID_WALLET_TYPE"),
+  INVALID_PASSWORD: t.literal("INVALID_PASSWORD"),
   WALLET_NOT_FOUND: t.literal("WALLET_NOT_FOUND"),
   INSUFFICIENT_PERMISSIONS: t.literal("INSUFFICIENT_PERMISSIONS"),
   GENERIC_IPC_ERROR
+}
+
+export const getWalletErrorMessages = {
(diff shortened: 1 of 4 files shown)


aesedepece:
This PR in its current form would still hide at least three different errors behind one single error code or message.

The point of issue #364 is precisely telling those each of those errors apart.

Namely, the errors are:
- The wallet does not exist. Indeed, we have never checked for this—it simply failed when getting the `wallet` entry.
- The password is wrong. This is easy to tell apart if we positively know that the wallet storage exists yet the `.get("wallet")` fails because of AES.
- The wallet is malformed. This is easy to spot too as it happens in the `asRuntimeType` call.

anler:
I have rolled back this PR and applied only the changes that fixes #364

anler:
I added yet another error type `CANNOT_ENCODE_WALLET`, at least if it fails we would know why, but I think this kind of error should be handled somewhere else

aesedepece:
Merged as 05d242766b108be808526d8b0fea6698eb1a41ea, 92831a6699999d07dd5e76327637daeaab354a9e and 70520ffcb99c723593d7448345d4ab3e06a72a75.

PR #500, opened 2018-09-04
Fix scrollbars

### LINKED ISSUE #489: Fix scrollbars
Regardless of the undeniable fact that in the technology preview we will have some suboptimal scrollbars, we still need to make sure that the elements that are currently overflowing and showing scrollbars are the right ones.

aesedepece:
Small changes to fix scrollbars.

A subsequent PR will greatly enhance responsiveness.

fix #489

### DIFF
--- a/app/renderer/ui/components/main/sections/wallet/tabs/send/index.tsx
+++ b/app/renderer/ui/components/main/sections/wallet/tabs/send/index.tsx
@@ -55,56 +55,58 @@ class TabSend extends TabComponent<any> {
     const selectFeesData: Array<SelectOptionData> = mockSelectFeesData
 
     return (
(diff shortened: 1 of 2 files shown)


app/renderer/ui/components/main/style.scss
```diff
@@ -25,18 +25,11 @@
     grid-template-areas:
       "tb"
       "ct";
-    grid-template-rows: repeat(2, 60px 1fr);
-    max-height: 100vh;
-    overflow: scroll;
+    grid-template-rows: 60px calc(100vh - 60px);
```
Tommytrg: Second row could be replaced by "1fr":
```
     grid-template-rows: 60px 1fr; 
```

PR #676, opened 2019-06-03
feat(ui-compo): implement Button Component

### LINKED ISSUE #610: Migrate Button component from legacy one
We have to migrate the default button component from legacy one. The expected result for this component should be similar to this:
photo5969778394688237323

kengork:
Close #610


src/components/Button.vue
```diff
@@ -0,0 +1,83 @@
...
+    onClick () {
+      this.$emit('click')
+    },
+  },
+}
+</script>
+
+<style lang="scss">
+@import '../styles/_colors.scss';
+button {
```
Tommytrg: Avoid non-class selectors

src/components/Button.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <button :class="type" @click="onClick">
```
Tommytrg: IMHO, we shouldn't pass the _type_ property to the class without check if it is a valid button type. What do you think about check if the type exists and if not render the default button?

src/stories/index.stories.js
```diff
@@ -3,23 +3,23 @@ import { storiesOf } from '@storybook/vue'
 import { action } from '@storybook/addon-actions'
 import { linkTo } from '@storybook/addon-links'
 
-import MyButton from '../components/MyButton.vue'
+import Button from '../components/Button.vue'
 
 storiesOf('Button', module)
   .add('with text', () => ({
-    components: { MyButton },
```
Tommytrg: We should add an instance of each button type

src/components/Button.vue
```diff
@@ -0,0 +1,121 @@
...
+  name: 'Button',
+  props: {
+    type: String,
+    nativeType: String,
+  },
+  methods: {
+    onClick () {
+      this.$emit('click')
+    },
+    checkType (propType) {
```
Tommytrg: I would suggest a more semantic name like _normalize_

src/components/Button.vue
```diff
@@ -0,0 +1,121 @@
...
+      switch (propType) {
```
Tommytrg: You can refactor this switch:
```
  switch (propType) {
    case 'default':
    case 'primary':
    case 'dashed':
    case 'danger':
    case 'positive':
       return propType
     default:
        return 'default'
  }
```

src/components/Button.vue
```diff
@@ -0,0 +1,121 @@
+<template>
+    <button :class="checkType(type)" @click="onClick" nativeType="submit">
```
Tommytrg: Add evaluation to _nativeType_ prop
```suggestion
  <button :class="checkType(type)" @click="onClick" :nativeType="nativeType">
```

src/components/Button.vue
```diff
@@ -0,0 +1,121 @@
...
+
+<script>
+export default {
+  name: 'Button',
+  props: {
+    type: String,
+    nativeType: String,
+  },
+  methods: {
+    onClick () {
```
Tommytrg: Remove method onClick and pass it as prop

PR #680, opened 2019-06-04
docs: fix documentation according the new boilerplate

Tommytrg:
This PR updates the documentation removing content about old features and adding those that are news

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,7 +13,7 @@ By participating, you are expected to uphold this code as well.
 ## I just have a question
 
 Please don't file an issue with questions.
-It's easier for you and for us if you go directly to [our Gitter chatroom][gitter], since it will keep our repositories clean and you will get a faster response.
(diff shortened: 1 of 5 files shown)


.github/CONTRIBUTING.md
```diff
@@ -93,22 +93,19 @@ Unsure where to begin contributing? You can start by looking through these good
 * Do not include issue numbers in the first line of your commit message
 * Reference issues and pull requests liberally after the first line
 * [Verify your commits][signing-commits] before creating a pull request
-* Standardize your commits:
-    - use `yarn commit` to write the commit message with the interactive guide
-    - clean up the problem/warnings from check before pushing your commits
+* Standardize your commits
```
kengork: I think we should remind the contributors what convention we actually use (changelog I think)

Tommytrg: That is explained in STYLEGUIDE.md, but I think it is a good idea to write an example of how to write a commit message:
```
feat(ui-compo): implement button component


test(button): fix button behavior
```

kengork: Cool!

PR #712, opened 2019-07-01
feat(ui-compo): add balances component

Tommytrg:
This PR includes:
  - [x] Balances component implementation
  - [x] Show balances stored in vuex
  - [x] Include Balances in Transaction tab

### DIFF
--- a/src/components/Transactions.vue
+++ b/src/components/Transactions.vue
@@ -1,16 +1,33 @@
 <template>
   <div class="transactions">
-    <TransactionList :transactions="transactions" />
+    <TransactionList class="list" :transactions="transactions" />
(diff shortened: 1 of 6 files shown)


src/components/Transactions.vue
```diff
@@ -1,16 +1,33 @@
...
 
 export default {
   name: 'Transactions',
   components: {
     TransactionList,
+    Balances,
+  },
+  computed: {
+    available () {
+      return this.$store.state.wallet.wallet.balances.available
```
aesedepece: What.

Tommytrg: This will be less verbose when we close #717

PR #743, opened 2019-07-24
feat: implement receive transaction

Tommytrg:
This PR implements the user flow to create an address to be able to receive funds. To achieve that we call the generate_address method and get_addresses to show the existing ones.


src/api.js
```diff
@@ -58,6 +58,7 @@ export class WalletApi {
   }
 
   _handleError (error) {
+    console.log('handle error', error)
```
gabaldon: Not sure if this console.log is useful

src/store/wallet.js
```diff
@@ -63,9 +63,11 @@ export default {
     setVTT (state, transaction) {
       state.vtt = transaction
     },
-
-    setGeneratedAddress (state, { address }) {
-      state.generatedAddress = address
+    setAddresses (state, { addresses }) {
```
gabaldon: It would be nice to have only one method that includes both. What do you think about changing it?

PR #775, opened 2019-08-02
Impl query stage component

Tommytrg:


src/components/RadonQuery.vue
```diff
@@ -1,286 +1,31 @@
...
-  data () {
-    return {
-      error: {
-        retrieve: false,
-        aggregate: false,
-        consensus: false,
-      },
-    }
+    RadonStage,
   },
```
gabaldon: It would be nice to have this method in the toolbar and delete this component so we can include these features directly in the editor.

PR #793, opened 2019-08-30
Radon to Markup

Tommytrg:


src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+  {
+    markup_type: 'option',
+    hierarchical_type: 'argument_option',
+    label: 'notBottom',
+    output_type: TYPES.BYTES,
+    meta: {},
+  },
+]
 
 export const OPERATOR_INFOS = {
```
tmpolaczyk: I would move all the operator data to a different file.

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+    getOr: [0x81, [PSEUDOTYPES.INNER]],
+    isOk: [0x82, [TYPES.BOOLEAN]],
+  },
+}
+
+function updateMarkup(markup, value) {
+  if (markup.markup_type === 'select') {
+    const optionsList = markup.options.map(option => option.label)
+
+    const selectedOperatorInfoEntry = Object.entries(OPERATOR_INFOS).find(operatorInfoEntry => {
```
tmpolaczyk: Using find is inefficient, you should build a map on startup and just index it by `(name, optionsList)` to get the desired `operatorId`, and then use the `operatorId` to index `OPERATOR_INFOS`.

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+      const optionsListCandidate = Object.values(typeSystem).find(x => {
+        return Object.values(x)
+          .map(y => y[0])
+          .includes(parseInt(operatorInfoEntry[0]))
+      })
+      // tienen el mismo nombre de operador
+      return name === value && areSoftEqualArrays(optionsList, Object.keys(optionsListCandidate))
+    })
+    const selectedOperatorInfo = selectedOperatorInfoEntry[1]
+    const outputType = Object.values(typeSystem).reduce((acc, item) => {
```
tmpolaczyk: Same, you can avoid the reduce/find by creating a helper map of `operatorId => outputType`.

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+  }
+
+  pushOperator(stage, sourceIndex) {
+    const script =
+      stage === 'retrieve' ? this.markup[stage][sourceIndex].script : this.markup[stage].script
+    const lastOperator = script[script.length - 1]
+    const lastOperatorOutputType = Array.isArray(lastOperator.selected.output_type)
+      ? lastOperator.selected.output_type[0]
+      : lastOperator.selected.output_type
+    const outputType = TYPES[lastOperatorOutputType.toUpperCase()]
```
tmpolaczyk: Isn't this a no-op?

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+    if (stage === 'retrieve') {
+      this.markup[stage][sourceIndex].script.push(newOperator)
+    } else {
+      this.markup[stage].script.push(newOperator)
+    }
+    return newOperator
+  }
+
+  updateElement(id, value, stage) {
+    const searchAndReplace = function(currentMarkup, id, value, found) {
```
tmpolaczyk: `found` is always false, and instead of setting `found` to true you can just not call the function.
```suggestion
    const searchAndReplace = function(currentMarkup, id, value) {
```

tmpolaczyk: And this entire function is very ineficient, you want a map of `id => markup`.

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+      return source
+    })
+
+    this.markup.aggregate.script = searchAndReplace(this.markup.aggregate.script, id, value, false)
+    this.markup.tally.script = searchAndReplace(this.markup.tally.script, id, value, false)
+  }
+
+  _getSourceMarkup(source) {
+    return source.map((operator, index) => {
+      return Object.entries(typeSystem).reduce((acc, entry) => {
```
tmpolaczyk: This is not a reduce

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+        // returns generateMarkdown([operatorName,[operatorCode, outputType] ])
+        const search = Object.entries(entry[1]).find(x =>
```
tmpolaczyk: Another find can can be removed

src/radon/index.js
```diff
@@ -1,761 +1,1181 @@
...
+                options: MARKUP_REDUCER_OPTIONS,
+              }
+            } else if (argument.type === TYPES.FILTER) {
+              return {
+                id: generateId(),
+                label: argument.name,
+                markup_type: 'input',
+                hierarchical_type: 'argument',
+                selected: Object.entries(FILTERS)
+                  .filter(filter => filter[1] === operator[index + 1])
```
tmpolaczyk: This is a find, which can be avoided by creating a map :)

PR #806, opened 2019-09-17
Generate address, list tx and vtt's

Tommytrg:


src/components/Send.vue
```diff
@@ -140,25 +168,29 @@ export default {
...
   .value {
     font-size: 16px;
     font-weight: 400;
   }
   .address {
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    width: 150px;
+    // overflow: hidden;
```
gabaldon: If this is not useful anymore, should it be deleted?

PR #857, opened 2019-12-23
Fix get addresses

gabaldon:
The method for setting the list of addresses when generating a new one has been divided into two different methods. The offset of addresses that can be listed is 25, there is a need for checking the argument `offset` to retrieve more than 25 addresses from the Wallet.

### DIFF
--- a/src/components/Transactions.vue
+++ b/src/components/Transactions.vue
@@ -73,13 +73,14 @@ export default {
     return {
       dialogVisible: false,
       dialogVisible2: false,
-      posts: this.addresses,
+      posts: this.newAddresses,
       lastAddress: '',
       page: 1,
       perPage: 5,
       middleItemActive: null,
       lastItemActive: false,
       firstItemActive: true,
+      newAddresses: null,
     }
   },
   computed: {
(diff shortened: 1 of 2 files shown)


src/components/Transactions.vue
```diff
@@ -100,7 +101,8 @@ export default {
         return state.wallet.errors.getTransactions
       },
       addresses: state => {
-        return state.wallet.addresses
+        const addressesList = Array.from(state.wallet.addresses)
```
Tommytrg: We can remove the return here

src/store/wallet.js
```diff
@@ -76,19 +76,16 @@ export default {
...
-    setAddresses(state, { addresses, address }) {
-      const addressesMap = new Map([])
-      const addAddress = addr => addressesMap.set(addr.address, addr)
-      state.addresses.map(addAddress)
-
-      if (addresses) {
-        addresses.map(addAddress)
-      }
-
+    setNewAddress(state, { address }) {
```
Tommytrg: What do you think about rename it as `addAddress`
```suggestion
    addAddress (state, { address }) {
```

PR #858, opened 2019-12-26
Add submit on pressing enter

gabaldon:
This PR adds a submit on pressing the enter key in the send component and in the unlock wallet component. This entails better keyboard accessibility.


src/components/WelcomeBack/UnlockWallet.vue
```diff
@@ -26,9 +30,13 @@ export default {
       password: '',
       showError: false,
       sent: false,
+      tabKeyPressed: false,
```
Tommytrg: I think you are not using this

PR #879, opened 2020-01-08
Tx subscription

### LINKED ISSUE #883: Update Sheikah after wallet resynch
If Sheikah is connected to a wallet and the wallet is stopped. When wallet is working again and Sheikah is also reconnected with it. From now to the future, transactions will not be updated in Sheikah.

I dont know if maybe it continues disconnected with wallet but show SYNC and the green circle, or it is connected but it doesnt update transactions anymore.

If you relaunch electron, it works again

Tommytrg:
We should make sure that subscriptions to new transactions that relate to a particular wallet do work.

Close #883

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -70,6 +70,13 @@ export class WalletApi {
         .catch(this._handleError)
   }
 
+  async subscribeToNotifications(params, cb) {
+    return this._callApiMethod('subscribeNotifications')([params.sessionId])
+      .then(_ => {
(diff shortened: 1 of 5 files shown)


aesedepece:
Are push subscriptions not possible? Polling is suboptimal IMO.

Tommytrg:
They are not possible but I think @mariocao and @girazoki will solve it.

I have added a subscription method that we'll call when the wallet supports it

PR #890, opened 2020-01-14
Marketplace view

gabaldon:
This PR implements a provisional view of the marketplace component.


src/api.js
```diff
@@ -148,7 +148,9 @@ export class MarketplaceApi {
   baseUrl = process.env.marketplaceUrl || 'https://witnet-marketplace-api-test.herokuapp.com'
 
   _handleResponse(response) {
-    return response && response.error ? response : response.data || true
+    if (response) {
```
Tommytrg:
```suggestion
    if (response && response.data) {
```

src/components/card/TemplateCard.vue
```diff
@@ -95,8 +121,13 @@ export default {
     name: String,
     description: String,
     date: [Number, String],
+    type: [Number, String],
   },
   methods: {
+    getClass(className) {
```
Tommytrg: You are always calling this method with `type prop`. Create a computed property instead

src/store/wallet.js
```diff
@@ -68,14 +68,15 @@ export default {
...
     },
-
     clearError(state, { error }) {
       state.errors[error] = null
     },
-
     setGeneratedTransaction(state, { transaction }) {
       state.generatedTransaction = transaction
     },
+    deleteGeneratedTransaction(state) {
```
Tommytrg:
```suggestion
    clearGeneratedTransaction(state) {
```

PR #892, opened 2020-01-15
Show errors and warnings

gabaldon:
This PR implements the error message when the password is not valid and a warning message when running Sheikah the Wallet is not synced.


src/router.js
```diff
@@ -22,25 +20,41 @@ import WalletSeedTypeSelection from '@/components/steps/WalletSeedTypeSelection.
...
+    next('/run-wallet-alert')
   }
 }
 
 export default new Router({
   routes: [
+    {
+      path: '/run-wallet-alert',
+      beforeEnter: redirectIfNeccesary,
+      name: 'runWalletAlert',
```
Tommytrg: What do you think about `walletNotFound` name or `MissingWallet`?

PR #926, opened 2020-02-11
Add standardize method to set currency unit

gabaldon:
This PR adds a method to standardize currency unit in balances and transactions.


src/utils.js
```diff
@@ -12,6 +12,22 @@ function encodeAggregationTally(stage) {
   }
 }
 
+export function standardizeWitUnits(amount, unit) {
+  const units = {
```
Tommytrg: Convert units into constants

PR #933, opened 2020-02-19
Sheikah bundle

Tommytrg:
Implement sheikah bundling process


package.json
```diff
@@ -33,7 +34,11 @@
     "cbor": "^4.1.5",
     "core-js": "^3.4.4",
     "element-ui": "^2.4.5",
+    "ps-list": "^7.0.0",
     "rpc-websockets": "^4.3.5",
+    "tar": "^6.0.1",
+    "tar-stream": "^2.1.0",
+    "targz": "^1.0.1",
```
tmpolaczyk: Why do you need 3 tar libraries?

src/background.js
```diff
@@ -86,3 +104,122 @@ if (isDevelopment) {
...
+        await pipeline(response.data, fs.createWriteStream(file))
+        console.log('witnet release downloaded succesfully')
+
+        console.log('Decompressing release...')
+        // Decompress tar.gz file
+        tar.x({
+          file,
+          sync: true,
+        })
+        // TODO: tar is not decompressing correctly if a pass a path.
```
tmpolaczyk: Maybe running

    exec(`cd ${SHEIKAH_PATH}`)

or equivalent before unpacking would help

src/background.js
```diff
@@ -86,3 +104,122 @@ if (isDevelopment) {
...
+  // Run by ./wintet => cmd: './witnet wallet server',
+  return !!processes
+    .filter(x => x.name === 'witnet')
+    .find(p => p.cmd.includes('/witnet wallet server'))
+}
+
+function runWallet() {
+  console.log('Running wallet...')
+  // Pass { detached: true, stdio: false } and then call .unref() in child_process will make it independent from its parent.
+  // In other words, if sheikah stops the wallet process will keep alive
```
tmpolaczyk: Is this a good idea? How do you stop the wallet? This will force users to run `killall witnet` if something goes wrong...

src/background.js
```diff
@@ -86,3 +104,122 @@ if (isDevelopment) {
...
+
+        resolve()
+      })
+      .catch(err => {
+        console.log('err', err)
+      })
+  })
+}
+
+async function isWalletRunning() {
```
tmpolaczyk: Why not just ping the wallet using JSON-RPC? You know the socket address, that way you don't need to parse the full list of processes.

src/background.js
Tommytrg: It hasn't worked

PR #934, opened 2020-02-24
Redirect when un-authorized user

gabaldon:
This PR adds redirection to the wallet list when the user is not authorized anymore.


src/store/wallet.js
```diff
@@ -109,6 +109,10 @@ export default {
         message,
       }
     },
+    clearSession(state) {
```
Tommytrg: I think this should be removed

PR #954, opened 2020-03-02
chore: retry curl if github is down

tmpolaczyk:
This will keep trying to download the witnet release until it works.

### DIFF
--- a/ci/initializeWallet.sh
+++ b/ci/initializeWallet.sh
@@ -2,7 +2,17 @@
 
 echo 'Downloading latest witnet node release...'
 
-wget $(curl -s https://api.github.com/repos/witnet/witnet-rust/releases/latest | grep 'browser_' | cut -d\" -f4)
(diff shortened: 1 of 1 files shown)


ci/initializeWallet.sh
```diff
@@ -2,7 +2,12 @@
 
 echo 'Downloading latest witnet node release...'
 
-wget $(curl -s https://api.github.com/repos/witnet/witnet-rust/releases/latest | grep 'browser_' | cut -d\" -f4)
+while URL=$(curl -s https://api.github.com/repos/witnet/witnet-rust/releases/latest | grep 'browser_' | cut -d\" -f4); [ "$URL" = "" ]; do
+    echo 'Github is down, retrying in 10s'
+    sleep 10s
+done
+
+wget $URL
```
Tommytrg: I think this `wget` could also fail

PR #958, opened 2020-03-04
Import wallet flow

gabaldon:
This PR implements the flow for importing a wallet except for the seed validation.

PR #969, opened 2020-03-11
Save and get claiming information + download file

gabaldon:
This PR save and get the claiming information from the wallet and download the JSON file with the information. It also creates an alternative 'create wallet' process for claiming tokens. 
It needs to validate the file and fill the steps with the correct disclaimers.


Tommytrg:
This PR is included in #958

PR #976, opened 2020-03-16
Show upload claiming file error

gabaldon:
This PR shows an error when no claiming file is uploaded and keep the claiming file available before it is saved in the wallet.

### DIFF
--- a/src/components/claiming/UploadFile.vue
+++ b/src/components/claiming/UploadFile.vue
@@ -21,12 +21,17 @@
         :http-request="handleUpload"
         :on-change="handleUpload"
         :limit="1"
+        :file-list="file"
         :on-exceed="handleExceed"
+        :on-remove="clearClaimingInfo"
       >
         <i class="el-icon-upload"></i>
-        <div class="el-upload__text">Suelta tu archivo aquí o <em>haz clic para cargar</em></div>
+        <div class="el-upload__text">Drag your file here or <em>click to upload</em></div>
         <div slot="tip" class="el-upload__tip">Only application/json files supported</div>
       </el-upload>
+      <p v-if="uploadFileError" class="error">
+        {{ uploadFileError.message }}
+      </p>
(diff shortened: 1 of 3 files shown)


src/components/claiming/UploadFile.vue
```diff
@@ -21,19 +21,25 @@
         :http-request="handleUpload"
         :on-change="handleUpload"
         :limit="1"
+        :file-list="file"
```
Tommytrg: Is there a way to change the UI when the file is uploaded?

src/components/claiming/UploadFile.vue
```diff
@@ -21,19 +21,25 @@
...
         :on-exceed="handleExceed"
+        :on-remove="clearClaimingInfo"
       >
         <i class="el-icon-upload"></i>
-        <div class="el-upload__text">Suelta tu archivo aquí o <em>haz clic para cargar</em></div>
+        <div class="el-upload__text">Drag your file here or <em>click to upload</em></div>
         <div slot="tip" class="el-upload__tip">Only application/json files supported</div>
       </el-upload>
+      <p v-if="uploadFileError" class="error">
```
Tommytrg: If you upload a file once and you try to upload it again it shows an error. The expected behaviour is allow upload the file as many times as you want

src/components/claiming/UploadFile.vue
gabaldon: It already changes. When the file is uploaded a green tick is shown to the user.

src/components/claiming/UploadFile.vue
gabaldon: You need to delete the current file and upload it again. You cannot upload more than one.

src/components/claiming/UploadFile.vue
gabaldon: Discuss in Issue #980

PR #992, opened 2020-03-18
feat: calculate addresses according to the number of wits

Tommytrg:
Calculate addresses according to the number of wits given.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,4 +1,4 @@
-export const CLAIM_WITS_PER_ADDRESS = 1
+export const CLAIMING_ADDRESS_MIN_WITS = 50
 
 export const WIT_UNIT = {
   WIT: 'wit',
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,6 @@
 import cbor from 'cbor'
 import uuidv4 from 'uuid/v4'
-import { WIT_UNIT } from '@/constants'
+import { WIT_UNIT, CLAIMING_ADDRESS_MIN_WITS } from '@/constants'
 
 // Create Notifications if notifications are supported
 export function createNotification(notificationProps) {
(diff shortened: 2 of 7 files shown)


src/utils.js
```diff
@@ -176,3 +176,25 @@ export function getNativeValueFromMarkupArgumentType(value, type) {
   }
   return value
 }
+
+// calculate the number of addresses according to the amount of wits passed
+export function calculateAddressesNumber(amount) {
+  if (amount === 0) return 0
+  const ceil50 = x => x + 50 - (x % 50)
```
tmpolaczyk: Ceil(50) should return 50, not 100

```suggestion
  const ceil50 = x => Math.ceil(x / 50) * 50
```

src/utils.js
```diff
@@ -176,3 +176,25 @@ export function getNativeValueFromMarkupArgumentType(value, type) {
...
+  // round the amount of wits for convenience and deniability
+  const roundedAmount = amount % 50 === 0 ? amount : ceil50(amount)
+  // exponents to calculate how many wits fit in an address
+  const arr = [6, 5, 4, 3, 2, 1, 0]
+  return arr.reduce(
```
tmpolaczyk:
```suggestion
  return (roundedAmount / 50).toString().split('').map(Number).reduce((a, b) => a + b, 0)
```

src/constants.js
```diff
@@ -1,4 +1,5 @@
 export const CLAIM_WITS_PER_ADDRESS = 1
+export const CLAIMING_MIN_WITS_ADDRESS = 50
```
tmpolaczyk: This name seems to imply that this is an address, because the unit is usually the last word. Maybe rename to:
```suggestion
export const CLAIMING_ADDRESS_MIN_WITS = 50
```

src/constants.js
```diff
@@ -1,4 +1,5 @@
 export const CLAIM_WITS_PER_ADDRESS = 1
```
tmpolaczyk: This is unused now?
```suggestion
```

PR #1003, opened 2020-03-23
Sign disclaimers

gabaldon:
This PR create disclaimers and signs them using the ```sign_data``` method provided by the wallet.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -70,67 +70,98 @@ export class WalletApi {
         .catch(this._handleError)
   }
 
-  async subscribeToNotifications(params, cb) {
+  subscribeToNotifications(params, cb) {
     return this._callApiMethod('subscribe_notifications')([params.sessionId])
       .then(_ => {
         this.client.on('notifications', cb)
       })
       .catch(this._handleError)
   }
-  async createDataRequest(params) {
+
+  createDataRequest(params) {
     return this._callApiMethod('create_data_request')(params)
   }
 
-  async createMnemonics(params) {
+  createMnemonics(params) {
     return this._callApiMethod('create_mnemonics')(params)
   }
 
-  async createWallet(params) {
+  createWallet(params) {
     return this._callApiMethod('create_wallet')(params)
   }
 
-  async generateAddress(params) {
+  signData(params) {
+    return this._callApiMethod('sign_data')(params)
+  }
+
+  // sign all disclaimers at once
+  signDisclaimers(params) {
+    const disclaimers = { disclaimer_1: 'disclaimer_1', disclaimer_2: 'disclaimer_2' }
+    const requests = []
+    Object.entries(disclaimers).forEach(elm => {
+      const request = this.signData({
+        wallet_id: params.wallet_id,
+        session_id: params.session_id,
+        data: elm[0],
+      })
+      requests.push(request)
+    })
+    return Promise.all(requests).then(request => {
+      const isError = request.find(x => !x.result)
+      if (isError) {
+        return { error: 'Error signing data' }
+      } else {
+        // assign disclaimer names to its signature
+        return Object.keys(disclaimers).reduce((acc, key, index) => {
+          acc[key] = request[index].result
+          return acc
+        }, {})
+      }
+    })
+  }
+
+  generateAddress(params) {
     return this._callApiMethod('generate_address')(params)
   }
 
-  async getTransactions(params) {
+  getTransactions(params) {
     return this._callApiMethod('get_transactions')(params)
   }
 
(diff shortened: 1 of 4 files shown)


src/api.js
```diff
@@ -89,6 +89,10 @@ export class WalletApi {
     return this._callApiMethod('create_wallet')(params)
   }
 
+  async signData(params) {
```
Tommytrg: This method doesn't need to be async

src/components/claiming/Loading.vue
```diff
@@ -32,13 +32,17 @@ export default {
...
+      disclaimers: state => state.wallet.disclaimers,
       error: state => state.wallet.errors.unlockWallet || state.wallet.errors.createWallet,
     }),
   },
   watch: {
     async sessionId(value) {
       if (value) {
         await sleep(2000)
+        Object.entries(this.disclaimers).forEach(elm => {
+          this.$store.dispatch('signData', { data: elm[0] })
```
Tommytrg: Instead of dispatch multiple times signData, I would dispatch it only once and send multiple sign data request from que vuex action.

src/components/claiming/WalletDisclaimer.vue
```diff
@@ -40,13 +40,15 @@ export default {
...
       texts: {},
+      disclaimer_2: 'Important',
     }
   },
   methods: {
     previousStep() {
       this.$router.push('/claiming/create-wallet')
     },
     nextStep() {
+      this.$store.commit('addDisclaimer', { key: this.disclaimer_2 })
```
Tommytrg: We don't need to add disclaimers in each step. I think we can assume that the user has accept of the disclaimers if they have create the wallet.

src/store/wallet.js
```diff
@@ -4,11 +4,13 @@ import { addToList, encodeDataRequest, standardizeWitUnits, createNotification }
 import { UPDATE_TEMPLATE } from '@/store/mutation-types'
 import { WIT_UNIT } from '@/constants'
 import warning from '@/resources/svg/warning.png'
+import { c } from 'tar'
```
Tommytrg: wrong editor autoimport?

src/store/wallet.js
```diff
@@ -46,6 +48,7 @@ export default {
     networkStatus: 'error',
     radRequestResult: null,
     transactions: [],
+    disclaimers: {},
```
Tommytrg: We don't need to store the raw disclaimers.They can be a constant and be used only in signData

src/store/wallet.js
```diff
@@ -493,6 +498,32 @@ export default {
...
+      const requests = []
+      Object.entries(disclaimers).forEach(elm => {
+        const request = context.state.api.signData({
+          wallet_id: context.state.walletId,
+          session_id: context.state.sessionId,
+          data: elm[0],
+        })
+        requests.push(request)
+      })
+      axios.all(requests).then(results => {
```
Tommytrg: Can you move that inside `api.js` with all the http calls?

src/store/wallet.js
```diff
@@ -493,6 +498,32 @@ export default {
...
+        results.forEach(res => {
+          if (res.result) {
```
Tommytrg: Here you can just `.find` an error and if the result is true, call `setError`

src/store/wallet.js
```diff
@@ -123,7 +126,9 @@ export default {
     },
     setClaimingInfo(state, { info }) {
       Object.assign(state, { claimingFileInfo: info })
-      console.log(state.claimingFileInfo)
+    },
+    setDisclaimers(state, { result }) {
+      state.disclaimers[result.signature] = result
```
Tommytrg: The keys of `disclaimers` should be the disclaimer name
```suggestion
    setDisclaimers(state, { result }) {
      state.disclaimers['disclaimer_1'] = result
```

src/store/wallet.js
```diff
@@ -123,7 +126,9 @@ export default {
     },
     setClaimingInfo(state, { info }) {
       Object.assign(state, { claimingFileInfo: info })
-      console.log(state.claimingFileInfo)
+    },
+    setDisclaimers(state, { result }) {
```
Tommytrg: You can pass all the signed disclaimers at once.

src/components/claiming/Loading.vue
```diff
@@ -5,6 +5,7 @@
       <Spinner class="spinner" :active="true" />
       <p class="description">This will take a few seconds</p>
     </div>
+    {{ error }}
```
Tommytrg: Remove log

PR #1006, opened 2020-03-25
Update claiming process theme

gabaldon:
This PR updates the claiming process theme to fit the new style.

### DIFF
--- a/src/components/claiming/WalletEncryptionPassword.vue
+++ b/src/components/claiming/WalletEncryptionPassword.vue
@@ -15,21 +15,25 @@
       phrase is still your ultimate backup.
     </p>
     <div class="form-row password">
-      <PasswordInput
+      <p>Create a password</p>
+      <el-input
         data-test="password-input"
-        label="Password"
+        placeholder="Please input password"
         v-model="password"
-        v-on:go-next="goNextInput"
-      />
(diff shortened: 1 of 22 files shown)


src/components/claiming/WalletEncryptionPassword.vue
```diff
@@ -15,21 +15,27 @@
...
+      <el-input
+        class="password"
         data-test="password-input"
-        label="Password"
+        placeholder="Please input password"
         v-model="password"
-        v-on:go-next="goNextInput"
-      />
+        show-password
+        @keydown.enter.native="goNextInput"
```
Tommytrg: Instead focus the next input on keydown enter should focus when pressing tab. I think you will find [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) useful

src/components/claiming/WalletEncryptionPassword.vue
```diff
@@ -73,7 +78,7 @@ export default {
       })
     },
     goNextInput() {
-      this.$refs.password.$refs.passInput.focus()
+      this.$refs.password.focus()
```
Tommytrg: You can remove that with the comment above

PR #1023, opened 2020-03-30
Disable development features in production environment

Tommytrg:


aesedepece:
Is this guarded in any way so that we can still use dev tooling bonanza while in development?

Tommytrg:
Shortcuts are only disable in development mode. On the other hand, we should never be able to go back because it's highly error prone.

aesedepece:
> Shortcuts are only disable in development mode. On the other hand, we should never be able to go back because it's highly error prone.

I guess you meant that they are only **enabled** in dev mode :stuck_out_tongue:

PR #1025, opened 2020-03-31
Fix standardizeWitUnits function

gabaldon:
This PR fixes the ```standardizeWitUnit``` function to not remove zero digits if the unit provided is nanoWit.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -38,7 +38,11 @@ export function standardizeWitUnits(amount, unit) {
     [`${WIT_UNIT.MICRO}`]: 3,
     [`${WIT_UNIT.NANO}`]: 0,
   }
-  return (amount / Math.pow(10, units[unit])).toFixed(units[unit]).replace(/\.?0+$/, '')
+  if (unit === WIT_UNIT.NANO) {
+    return amount.toString()
+  } else {
+    return (amount / Math.pow(10, units[unit])).toFixed(units[unit]).replace(/\.?0+$/, '')
+  }
 }
 
 export function encodeDataRequest(radRequest) {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -38,7 +38,11 @@ export function standardizeWitUnits(amount, unit) {
     [`${WIT_UNIT.MICRO}`]: 3,
     [`${WIT_UNIT.NANO}`]: 0,
   }
-  return (amount / Math.pow(10, units[unit])).toFixed(units[unit]).replace(/\.?0+$/, '')
+  if (unit === WIT_UNIT.NANO) {
+    return (amount / Math.pow(10, units[unit])).toFixed(units[unit])
```
aesedepece: Hey I think the division in line 42 is redundant as `units[WIT_UNIT.NANO]` always equals `0` and `Math.pow(10, 0)` is `1`.

Same happens with `.toFixed(units[unit])`, it does nothing more than `.toFixed()` with no arguments, or simply `.toString()` :wink:

src/utils.js
```diff
@@ -38,7 +38,11 @@ export function standardizeWitUnits(amount, unit) {
     [`${WIT_UNIT.MICRO}`]: 3,
     [`${WIT_UNIT.NANO}`]: 0,
   }
-  return (amount / Math.pow(10, units[unit])).toFixed(units[unit]).replace(/\.?0+$/, '')
+  if (unit === WIT_UNIT.NANO) {
+    return amount
```
aesedepece, line 42: It looks a little strange to me that one branch here returns `Number` while the other returns `String`.

Most likely it's not going to be a problem in our case, but I'd sleep better at night with this:

```suggestion
    return amount.toString()
```

Sorry for the nitpicking :sweat_smile:

PR #1051, opened 2020-04-15
Update side bar with grid

gabaldon:
This PR updates the sidebar to match the new style and uses ```grid``` for the layout.


src/components/DotIndicator.vue
```diff
@@ -24,10 +25,14 @@ export default {
...
+  border: 2px solid $green-5;
+  border-radius: 4px;
 }
 
 .error {
-  color: $red-5;
+  width: 30px;
+  border: 2px solid $red-5;
+  border-radius: 4px;
 }
```
aesedepece, line 37: We will now also have a `syncing` status. Could be `#EEBA1B` for example.

src/components/NetworkStatus.vue
```diff
@@ -1,9 +1,42 @@
...
+          v-else
+          src="@/resources/svg/short-down.svg"
+          @click="showAll = true"
+          alt="sort-down"
+        />
+      </div>
+    </div>
+    <div v-if="showAll && hover" class="detail-info">
+      <p class="text">
+        Connected to <span class="bold"> {{ node }} </span>
```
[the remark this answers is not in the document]
Tommytrg: I agree with that.

src/components/Sidebar.vue
```diff
@@ -1,56 +1,59 @@
 <template>
   <div
-    @mouseover="hover = true"
-    @mouseleave="hover = false"
-    :class="[hover ? 'sidebar' : 'sidebar collapsed-sidebar']"
+    @mouseover="expanded = true"
+    @mouseleave="expanded = false"
+    :class="[expanded ? 'sidebar expanded' : 'sidebar collapsed-sidebar']"
```
aesedepece, line 5: I think this is possible:
```suggestion
    :class="['sidebar', expanded ? 'expanded' : 'collapsed-sidebar']"
```

src/components/Sidebar.vue
```diff
@@ -1,56 +1,59 @@
...
   >
-    <div :class="[hover ? 'brand' : 'brand collapsed-brand']">
-      <router-link data-test="logo-to-main" class="logo" to="/wallet/transactions">
-        <img v-if="hover" class="sheikah-img" src="@/resources/svg/sheikah.svg" />
-        <img v-else class="sheikah-img" src="@/resources/svg/sheikah-small.svg" />
-      </router-link>
-    </div>
+    <router-link
+      data-test="logo-to-main"
+      :class="[expanded ? 'brand expanded' : 'brand collapsed-brand']"
```
aesedepece, line 9:
```suggestion
      :class="['brand', expanded ? 'expanded' : 'collapsed-brand']"
```

src/components/Sidebar.vue
```diff
@@ -1,56 +1,59 @@
...
       <router-link data-test="to-community" class="link" to="/community">
         <font-awesome-icon class="icon" icon="users" />
-        <span :class="[hover ? 'label' : 'label collapsed-label']">Community</span>
+        <span v-if="expanded" class="label">Community</span>
       </router-link>
     </div>
-    <div :class="[hover ? 'settings' : 'settings collapsed-settings']">
-      <Settings :settings="settings" color="dark" />
-      <NetworkStatus :windowWidth="windowWidth" :hover="hover" :status="status" />
+    <div :class="[expanded ? 'settings expanded' : 'settings collapsed-settings']" @click="closeSession()">
```
aesedepece, line 46:
```suggestion
    <div :class="['settings', expanded ? 'expanded' : 'collapsed-settings']" @click="closeSession()">
```

src/components/Sidebar.vue
```diff
@@ -1,56 +1,59 @@
...
       <router-link data-test="to-marketplace" class="link" to="/marketplace">
         <font-awesome-icon class="icon" icon="shopping-bag" />
-        <span :class="[hover ? 'label' : 'label collapsed-label']">Marketplace</span>
+        <span v-if="expanded" class="label">Marketplace</span>
       </router-link>
 
       <router-link data-test="to-community" class="link" to="/community">
         <font-awesome-icon class="icon" icon="users" />
-        <span :class="[hover ? 'label' : 'label collapsed-label']">Community</span>
+        <span v-if="expanded" class="label">Community</span>
```
aesedepece, line 43: I wonder whether removing the `<span>` elements is necessary if the sidebar itself is being resized down when collapsed, so they will be hidden.

In any case, I think `v-show` could be better than `v-if` in this case as it doesn't cause the element to be mounted and unmounted from the DOM but only toggles the `display` CSS property.

src/components/NetworkStatus.vue
```diff
@@ -28,28 +73,65 @@ export default {
...
-  display: flex;
-  align-items: center;
-
-  .label {
-    min-width: max-content;
-    color: $sidebar-mainnet-color;
-    font-weight: $sidebar-mainnet-font_weight;
+  display: grid;
+  grid-template-rows: minmax(100px, max-content);
+  grid-template-columns: 350px;
```
aesedepece, line 78: I think I need a Tech Friday on CSS grids...

PR #1052, opened 2020-04-16
New addresses component

Tommytrg:
Apply new design of the addresses component and the necessary components to build it.


src/components/AddressInformation.vue
```diff
@@ -89,6 +91,14 @@ export default {
       copied: false,
     }
   },
+  computed: {
+    blindPkh() {
+      const start = this.pkh.slice(0, 20)
+      const end = this.pkh.slice(29)
```
aesedepece, line 97:
```suggestion
      const end = this.pkh.slice(-16)
```

src/components/AddressInformation.vue
```diff
@@ -0,0 +1,221 @@
...
+    font-style: italic;
+    line-height: 150%;
+
+    .link {
+      font-weight: 500;
+      color: $link-grey;
+      text-decoration: underline;
+    }
+  }
+}
```
aesedepece, line 220: Make sure the margins of this element match de design :wink:

src/components/Addresses.vue
```diff
@@ -0,0 +1,112 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/_colors.scss';
+
+.card {
+  border: 1px solid $grey-1;
+  border-radius: 4px;
+  /* height: 370px; */
```
gabaldon, line 80: Add  ```overflow: hidden;``` to avoid border-radius clip the content.

aesedepece: I was about to comment exactly the same hahaha

aesedepece:
Please use `Roboto Mono` for addresses. We could even make it a default for `pre` elements.

aesedepece:
The padding of the used / not used label look slightly off to me:

image

I think the card could also use some more bottom padding, but I haven't measured if it's the same as in figma, so it could be also wrong in the design.

aesedepece:
Just a reminder that we need to come with a better empty state:
image

Tommytrg:
> Just a reminder that we need to come with a better empty state (specially `#-1` doesn't make much sense hahaha):
> 
> image

I didn't think in the empty state assuming the wallet always generates at least one address. If it isn't the case, I will add an empty state.

aesedepece:
> I didn't think in the empty state assuming the wallet always generates at least one address. If it isn't the case, I will add an empty state.

Yeah let's have an empty state until we make sure the assumption from the wallet becomes true. Sorry if a few days ago I provided confusing feedback about this :wink:

aesedepece:
When you have two rows of addresses, the vertical space between the rows is excessive. Funnily enough, when a third row is added, everything looks just fine.

aesedepece:
I guess it is a known limitation, but this component will not show more than 25 addresses at the same time (probably the issue is not in the component itself but in `getAddresses`).

mariocao:
Yesterday I checked the code for `getAddresses` and it seemed to the same as `getTransactions`. I will double-check it again. :)

PR #1068, opened 2020-04-24
fix: style in templates and editor view

gabaldon:
Fix small style issues in templates and editor view.


src/utils.js
```diff
@@ -11,6 +11,11 @@ export function createNotification(notificationProps) {
   }
 }
 
+// crop string to selected number of caracters
+export function resize(address, caracters) {
+  return address.length > caracters ? `${address.substring(0, caracters)}...` : address
+}
+
```
Tommytrg, line 18:
```suggestion
// crop string to selected number of caracters
export function resize(address, characters) {
  return address.length > characters ? `${address.substring(0, characters)}...` : address
}

```

PR #1071, opened 2020-04-27
feat: validate funds before create vtt

### LINKED ISSUE #1072: Validate available funds before create a VTT
Validate available funds before create a value transfer transaction

Tommytrg:
Validate available funds before create a value transfer transaction

Close #1072

### DIFF
--- a/src/components/Send.vue
+++ b/src/components/Send.vue
@@ -128,7 +128,7 @@
             @keydown.enter.esc.prevent="createVTT"
             type="primary"
             data-test="sign-send-btn"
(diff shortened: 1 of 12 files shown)


src/components/Send.vue
```diff
@@ -175,6 +185,10 @@ export default {
   },
   computed: {
     ...mapState({
+      availableBalance: state => {
+        // TODO: change for avaiable when wallet return it
```
aesedepece, line 189:
```suggestion
        // TODO: change for available when wallet returns it
```

PR #1074, opened 2020-04-28
Fix error listing addresses

Tommytrg:


src/components/Addresses.vue
```diff
@@ -49,7 +49,7 @@ export default {
   },
   data() {
     return {
-      selectedIndex: this.addresses.length - 1 || null,
+      selectedIndex: this.addresses.length - 1 >= 0 ? this.addresses.length - 1 : -1,
```
aesedepece, line 52: The expression `this.addresses.length` does always fulfill the `>= 0` requirement, while the base value is `-1`, so the whole expression actually equals to this:
```suggestion
      selectedIndex: this.addresses.length - 1,
```

src/components/Send.vue
```diff
@@ -148,6 +148,15 @@ export default {
     Alert,
   },
   data() {
+    const enoughFunds = (rule, value, callback) => {
```
aesedepece, line 151: What's `rule` here?

src/components/Send.vue
```diff
@@ -175,6 +185,10 @@ export default {
   },
   computed: {
     ...mapState({
+      availableBalance: state => {
+        // TODO: change for avaiable when wallet return it
```
aesedepece, line 189:
```suggestion
        // TODO: change for available when wallet returns it
```

tests/unit/src/components/Addresses.spec.js
```diff
@@ -81,6 +81,24 @@ describe('Addresses.vue', () => {
         }),
       })
 
+      expect(wrapper.find(AddressInformation).exists()).toBe(true)
+    })
+    it("should render properly if there aren't addresses", () => {
```
aesedepece, line 86:
```suggestion
    it("should render an empty state if there aren't addresses", () => {
```

src/components/Send.vue
Tommytrg: This function is used by `Element` to define [custom validation rules](https://element.eleme.io/#/en-US/component/form#custom-validation-rules). `rule` contains information about the validations that you specified in the template such as the type or the the validator function for that type.

src/components/Addresses.vue
Tommytrg: Ashamed :no_mouth:

Tommytrg:
This is already fixed

PR #1076, opened 2020-04-28
Connect address components with wallet

Tommytrg:
- Connect address components with wallet
- Allow add a normalizer function on api calls.


src/components/Addresses.vue
```diff
@@ -7,16 +7,17 @@
         v-on:generate-address="$emit('generate-address')"
         v-on:select-address="selectAddress"
       />
+      {{ selectedAddress }}
       <AddressInformation
         v-if="selectedAddress"
         :pkh="selectedAddress.pkh"
```
gabaldon, line 13: You could use ```cropString``` from utils for selected addresses

PR #1089, opened 2020-05-04
Create modal for renaming templates

gabaldon:
This PR creates a modal for renaming templates and change the template description.

### DIFF
--- a/src/components/RenameTemplate.vue
+++ b/src/components/RenameTemplate.vue
@@ -1,0 +1,75 @@
+<template>
+  <el-dialog title="Rename template" :visible="true" width="30%">
+    <div class="info">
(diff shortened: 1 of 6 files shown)


src/components/RenameTemplate.vue
```diff
@@ -0,0 +1,75 @@
...
+  computed: {
+    style() {
+      return this.type
+    },
+    updateName: {
+      get() {
+        return this.name
+      },
+      set(newName) {
+        this.changeTemplateName({ id: this.id, name: newName })
```
Tommytrg, line 40: For the sake of simplicity I would put all the logic related with save template info in this component and emit the new template information on update.

PR #1100, opened 2020-05-05
Editor toolbar

Tommytrg:


src/components/EditorToolBar.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div class="toolbar">
+    <div class="left">
+      <a :href="dataStr" ref="download" download="template.json" style="display:none"></a>
+
+      <router-link data-test="go-back-to-templates" class="back" to="/request/templates">
+        <el-button class="back-btn" type="primary"><font-awesome-icon class="content" icon="angle-left" /></el-button>
```
gabaldon, line 7: This button could be just a router-link, the border of a primary button is rounded when it is hovered or activated and we don´t want that behaviour.

PR #1101, opened 2020-05-05
Add notification when a transaction is sent successfully

gabaldon:
This PR adds a notification when a transaction is sent successfully

### DIFF
--- a/src/store/wallet.js
+++ b/src/store/wallet.js
@@ -245,7 +245,6 @@ export default {
         console.log('----- Template deployed successfully -----')
         context.dispatch('saveLabel', { label, transaction: transactionToSend })
         context.commit('clearGeneratedTransaction')
-        console.log(transactionToSend)
         createNotification({
           title: 'Transaction succesfully sent',
           body: `The transaction ${cropString(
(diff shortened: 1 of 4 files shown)


src/store/wallet.js
```diff
@@ -245,6 +245,15 @@ export default {
...
         context.dispatch('saveLabel', { label, transaction: transactionToSend })
         context.commit('clearGeneratedTransaction')
+        console.log(transactionToSend)
+        createNotification({
+          title: 'Transaction succesfully sent',
+          body: `The transaction ${cropString(
+            transactionToSend.transaction_id,
+            12,
+            'middle'
+          )} has been sent succesfully`,
```
aesedepece, line 255:
```suggestion
          )} has been sent succesfully into the Witnet network.\nIt should be written into a block soon.`,
```

src/store/wallet.js
```diff
@@ -245,6 +245,15 @@ export default {
         console.log('----- Template deployed successfully -----')
         context.dispatch('saveLabel', { label, transaction: transactionToSend })
         context.commit('clearGeneratedTransaction')
+        console.log(transactionToSend)
```
Tommytrg, line 248: Remove this log

PR #1123, opened 2020-05-18
fix: update template listing order

Tommytrg:


src/components/Templates.vue
```diff
@@ -94,7 +95,7 @@ export default {
               ...template[1],
             }
           })
-          .sort((a, b) => parseInt(a.creationDate) - parseInt(b.creationDate)),
+          .sort((a, b) => parseInt(b.creationDate) - parseInt(a.creationDate)),
```
gabaldon, line 98: I think the order we discussed was last one edited first right?

src/components/Templates.vue
```diff
@@ -94,7 +95,11 @@ export default {
               ...template[1],
             }
           })
-          .sort((a, b) => parseInt(a.creationDate) - parseInt(b.creationDate)),
+          .sort((a, b) => {
+            console.log('a', a.lastTimeOpened)
```
gabaldon, line 99: could you remove this console.log?

PR #1136, opened 2020-05-21
fix(transaction): set correct origin addresses

Tommytrg:


src/components/Transaction.vue
```diff
@@ -93,7 +93,9 @@ export default {
   },
   computed: {
     address() {
-      return this.inputs.length > 1 ? 'several addresses' : this.inputs[0].address
+      const addressOrigin = this.type === 'POSITIVE' ? 'inputs' : 'outputs'
+
+      return this[addressOrigin].length > 1 ? 'several addresses' : this[addressOrigin][0].address
```
tmpolaczyk, line 98: This will not work when sending transactions to someone else, because almost always there will be an additional change output. Ideally change outputs would be marked somehow, but for now a best-effort approach would be to show the first output address as "To" and assume the second one is always the change.

tests/unit/src/components/Transaction.spec.js
```diff
@@ -62,109 +62,271 @@ describe('Renders the correct elements when click is not triggered', () => {
...
+
+    it('should finds negative or positive icon', () => {
+      expect(wrapper.contains('[data-test="negative-positive"]')).toBe(true)
+    })
+
+    it('should not find the origin element', () => {
+      expect(wrapper.contains('[data-test="origin"]')).toBe(false)
+    })
+
+    it('should find the address', () => {
```
tmpolaczyk, line 164: ?
```suggestion
    it('should not find the address', () => {
```

PR #1140, opened 2020-05-22
Implement aggregate and tally stage in Editor

gabaldon:
Implements aggregate and tally stage in Editor. It includes filters and the reducer for each step.


src/components/OperatorSelect.vue
```diff
@@ -0,0 +1,30 @@
+{ filters: [ { "hierarchicalType": "operator", "id": 5, "label": "deviationAbsolute", "markupType":
```
Tommytrg, line 1: We are going to parse this to be able to render this component :rofl:

PR #1141, opened 2020-05-22
fix: disable shortcuts using hidden menu

Tommytrg:
Disable shortcuts using a hidden menu binding the shortcut we want to disable to an option. This aims to remove the globalShortcut problems.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -8,11 +8,11 @@ import util from 'util'
 import { spawn } from 'child_process'
 import axios from 'axios'
 import tar from 'tar'
-import { BrowserWindow, app, protocol, globalShortcut } from 'electron'
 import {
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -102,9 +117,6 @@ app.on('ready', async () => {
       console.error('Vue Devtools failed to install:', e.toString())
     }
   } else {
-    globalShortcut.register('CmdOrCtrl+R', () => {})
-    globalShortcut.register('CmdOrCtrl+Shift+R', () => {})
-    globalShortcut.register('CmdOrCtrl+Shift+I', () => {})
   }
```
tmpolaczyk, line 120: I guess this else is not needed now, as it's empty

PR #1179, opened 2020-06-04
Remove import from xprv

gabaldon:
Fixes update of current wallet in wallet list select and removes import from xprv.


src/components/WelcomeBack/WalletList.vue
```diff
@@ -91,6 +91,9 @@ export default {
     },
   },
   watch: {
+    wallets() {
```
Tommytrg, line 94: How many times is this going to be called?

PR #1181, opened 2020-06-04
fix(transactions): show pagination only when required

Tommytrg:
Show transactions list pagination only when the total amount of transactions is greater than the items per page constant.

- [x] Add tests


src/components/TransactionList.vue
```diff
@@ -140,3 +150,885 @@ export default {
...
+
+<docs>
+### Examples
+#### Empty
+```jsx
+  <TransactionList :transactions="[]" currency="nanoWits" />
+```
+#### With pagination
+```jsx
+  const transactions = [
```
tmpolaczyk, line 162: It's generally faster to use `JSON.parse` when creating large objects in javascript. Also it would be great if the transaction was imported from somewhere, instead of polluting the component.

Tommytrg: The main goal of these `docs` is show how to use a specific component. IMO, If we didn't show the props passed to the component, this `doc` would be worthless.

Tommytrg: On the other hand, I think the length should be reduced.

src/store/wallet.js
```diff
@@ -485,6 +485,8 @@ export default {
         page,
       })
 
+      console.log(request.result)
```
gabaldon, line 488: Remove console.log?

PR #1205, opened 2020-06-15
Try data request when the request is updated

### LINKED ISSUE #1198: Avoid remove the result of the Try data request when the operator is added

gabaldon:
Try the data request automatically when it is updated and add a loader.
close #1198


gabaldon:
close #1198

src/components/DotsLoading.vue
```diff
@@ -21,6 +25,14 @@ export default {
       type: String,
       default: '4px',
     },
+    margin: {
+      type: String,
+      default: '8px 8px 0 8px',
```
Tommytrg, line 30: Could this default value be placed in styles?

PR #1209, opened 2020-06-16
Fix error result messages

### LINKED ISSUE #1200: Fix result messages in editor
Filter Radon Error and a format the string that is going to be shown

gabaldon:
It Fixes the error result message that is shown in partial results to be more readable. 

close #1200

### DIFF
--- a/src/components/RadonOperator.vue
+++ b/src/components/RadonOperator.vue
@@ -78,6 +78,7 @@ import {
   standardizeOperatorName,
   getNativeValueFromMarkupArgumentType,
   standardizeOutputType,
+  selectInnerError,
 } from '@/utils'
 import {
   UPDATE_TEMPLATE,
@@ -135,7 +136,7 @@ export default {
     }),
     radonError() {
       return this.outputLabel === 'error' || !this.operatorOutput
-        ? this.error
+        ? selectInnerError(this.error)
         : null
     },
     isReducer() {
--- a/src/utils.js
+++ b/src/utils.js
@@ -153,6 +153,15 @@ export function isRepeated(key, list) {
   return list && list[key]
 }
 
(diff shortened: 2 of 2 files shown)


src/components/RadonOperator.vue
```diff
@@ -128,8 +129,9 @@ export default {
       variables: state => state.rad.currentTemplate.variables,
     }),
     radonError() {
+      console.log(this.error)
```
aesedepece, line 132: Is this `console.log` intentional?

In any case, we have `console.error` now :fireworks: 
```suggestion
      console.error(this.error)
```

gabaldon: It should be removed anyway, but thanks!

src/utils.js
```diff
@@ -145,6 +145,19 @@ export function isRepeated(key, list) {
...
+// Take the substring between two strings
+export function selectInnerError(string) {
+  if (string) {
+    const substringInBetweenRegex = /(?<=Inner: `Some\()(.*)(?=\))/
+    const matchedArray = string.match(substringInBetweenRegex)
+    if (matchedArray) {
+      return matchedArray[0]
+    } else {
+      return string
+    }
```
Tommytrg, line 157: Personal opinion

```suggestion
    return matchedArray ? matchedArray[0] : string
```

src/utils.js
```diff
@@ -145,6 +145,19 @@ export function isRepeated(key, list) {
   return list && list[key]
 }
 
+// Take the substring between two strings
+export function selectInnerError(string) {
+  if (string) {
+    const substringInBetweenRegex = /(?<=Inner: `Some\()(.*)(?=\))/
```
aesedepece, line 151: #RegexPride

PR #1210, opened 2020-06-16
Disable protocol and content type options

### LINKED ISSUE #1196: Disable protocol and content-type selector
Only allow "JSON API" and "HTTP-GET"

gabaldon:
Remove protocol and content type options.

close #1196

### DIFF
--- a/src/components/EditorSource.vue
+++ b/src/components/EditorSource.vue
@@ -5,6 +5,7 @@
         <label class="label">Protocol</label>
         <Select
           v-model="localProtocol"
+          :disabled="true"
           :options="[{ primaryText: 'HTTP-GET' }]"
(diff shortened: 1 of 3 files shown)


src/components/EditorSource.vue
```diff
@@ -67,10 +69,11 @@ export default {
...
      */
-    contentType: {
-      required: true,
-      type: String,
-    },
+    // FIXME: Support other content types and protocol options
+    // contentType: {
+    //   required: true,
+    //   type: String,
+    // },
```
Tommytrg, line 76: `contentType` should be a prop in spite of we only allow one value. Doing that, when we support more contentTypes, we only have to remove the `disabled option`

PR #1226, opened 2020-06-30
Fix tooltip

clbartoli:
This PR fixes the tooltips so that the comment appears in the corresponding button. In particular fixes:
- undo/redo changes
- generate new address

TODO:
fix the copy address tooltip


gabaldon:
Duplicated in #1227

Issue #1234, opened 2020-07-13
Prevent result to show an error in previous operators if it is not necessary


gabaldon:
Prevent to show an error in previous operators when there is an error in the last default operator added.

Tommytrg:
@gabaldon Could you provide an example to replicate it?

PR #1242, opened 2020-07-16
Show a notification when there is a general error trying the data request

gabaldon:
This pr shows a notification when the dr is broken and there are no partial results as a result of trying the data request.

### DIFF
--- a/src/store/wallet.js
+++ b/src/store/wallet.js
@@ -7,6 +7,7 @@ import {
   encodeDataRequest,
   isSyncEvent,
   standardizeWitUnits,
+  selectInnerError,
 } from '@/utils'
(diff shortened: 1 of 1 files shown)


src/store/wallet.js
```diff
@@ -115,7 +116,19 @@ export default {
       }
     },
     setDataRequestResult(state, { result }) {
-      state.radRequestResult = { ...result, timestamp: Date.now() }
+      const RetrievePartialResults = result.result.retrieve.map(
```
Tommytrg, line 119:
```suggestion
      const retrievePartialResults = result.result.retrieve.map(
```

Tommytrg:
This PR improve user experience showing the user when an non-operator related error ocurred but it doesnt close #1234

PR #1252, opened 2020-07-20
feat(build): disallow copy mnemonics

### LINKED ISSUE #1249: Disallow copy mnemonics in production environment
Disallow select the mnemonics while the user is creating a new wallet to avoid copy them.

Tommytrg:
close #1249


src/components/steps/WalletSeedBackup.vue
```diff
@@ -35,6 +35,9 @@ export default {
     ...mapState({
       seed: state => state.wallet.mnemonics,
     }),
+    allowCopyMnemonics () {
+      return process.env.NODE_ENV === 'development'
+    }
```
aesedepece, line 40: Can we make this computed property and the CSS class consistent in their naming so that we don't need to negate one?

PR #1287, opened 2020-07-31
feat: replace try data request btn with switch

### LINKED ISSUE #1243: Add switch to toggle auto try data request
The data request editor toolbar contain "Try data request" button spite of auto try the data request the user is composing. Because of that, the editor should have a switch to be able to toggle the auto try data request instead of a button.

Tommytrg:
- Switch button color should be added
- Add and fix tests

close #1243


Tommytrg:
Included in #1288

PR #1289, opened 2020-08-03
Fix update when variable is connected to an input

gabaldon:


src/components/RadonOperator.vue
```diff
@@ -264,7 +265,6 @@ export default {
           id,
           value: getNativeValueFromMarkupArgumentType(value, type),
         })
-        this.tryDataRequest()
```
Tommytrg, line 267: How this fixes update a variable when is connected to an input and continues updating the partial result?

Tommytrg: This `tryDataRequest` is redundant. `tryDataRequest` should be called in vuex only

PR #1307, opened 2020-08-18
Fix map arguments to work without quotes

Tommytrg:


package.json
```diff
@@ -55,7 +55,7 @@
     "vue-json-viewer": "^2.2.13",
     "vue-router": "^3.0.3",
     "vuex": "^3.0.1",
-    "witnet-radon-js": "0.7.3"
+    "witnet-radon-js": "../witnet-radon-js"
```
Tommytrg, line 58: This will be replaced with the new witnet-radon-js version

gabaldon (approved):

PR #1313, opened 2020-08-25
feat(Balances): add tooltip explaining locked wits

### LINKED ISSUE #1310: Add informative icon in locked balance
After the claimming process, some users will have locked balances in their wallets. In order to improve their user experience, locked balances should have an icon which displays a popover on hover with the locked information.

Tommytrg:
Do you have suggestions about the style or the copy? 

image
image

Close #1310


src/components/BalanceData.vue
```diff
@@ -56,12 +72,36 @@ export default {
...
+      return EXTERNAL_URL.BALANCE_LOCKED
+    },
+  },
 }
 </script>
 
 <style lang="scss" scoped>
 @import '@/styles/_colors.scss';
 
+.info-message {
```
gabaldon, line 90: The color of the text inside the popover should be $alt-grey-5

gabaldon (changes requested):

gabaldon (approved):

PR #1348, opened 2020-09-14
Fix redirection when wallet infos are found and time to open the wallet

### LINKED ISSUE #1340: Fix time to open wallet when validating password
The time to show the wallet when writing the password is longer than expected. This should be fixed.

gabaldon:
This pr allows opening the wallet right after the session id is found and also fixes the redirection when the wallet info is found.
close #1340

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -54,11 +54,8 @@ export default new Router({
       path: '/',
       name: 'main',
       component: Main,
-      beforeEnter: (to, from, next) => {
+      beforeEnter: async (to, from, next) => {
         const isReady = store.state.wallet.api.client.ws.ready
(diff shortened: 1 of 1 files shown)


src/router.js
```diff
@@ -105,8 +105,10 @@ export default new Router({
           }
         } else {
           if (isReady) {
-            store.dispatch('getWalletInfos')
-            if (sessionId) {
+            await store.dispatch('getWalletInfos')
+            const isSessionId = store.state.wallet.sessionId
+            const walletInfos = store.state.wallet.walletInfos
```
Tommytrg, line 110: Could you avoid re-declare this?

Tommytrg (approved):

PR #1386, opened 2020-09-29
fix(claiming-process): transform nanowits from claiming in wits

### LINKED ISSUE #1384: Handle claiming file tokens as nanowits
Hereinafter, the claiming file includes tokens as nanowits instead of wits. Sheikah should handle them according to that

Tommytrg:
Close #1384

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -123,8 +123,6 @@ ipcMain.on('shutdown-finished', () => {
 
 // Exit cleanly on request from parent process in development mode.
 if (isDevelopment) {
-  main()
-
   if (process.platform === 'win32') {
     process.on('message', data => {
       if (data === 'graceful-exit') {
--- a/claiming_file_success.json
+++ b/claiming_file_success.json
@@ -1,16 +1,17 @@
 {
   "data": {
-    "email_address": "johndoe@example.tld",
-    "name": "John Doe",
-    "usd": 1000,
-    "wit": 99000,
-    "genesis_date": 1587081600,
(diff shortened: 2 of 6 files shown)


aesedepece:
My point is that everything not facing the user should be dealt as nanowits, and only convert to wits in the UI.

aesedepece:
I think we can truncate to 2 decimal positions for the sake of legibility:
`4,102.72,257,625` vs. `4,102.72`

aesedepece:
Also, numbers like `67556739.0241109` are being formated like `67,556,739.0,241,109`, which has wrong grouping on the decimal side.

aesedepece:
This is still generating addresses of amount `50`, while the smallest possible value should now be `50_000_000_000`.

aesedepece:
Specifically this line:
https://github.com/witnet/sheikah/blob/9e6588e22c729c1ce8dfab190cac9d3eac7c7b1f/src/utils.js#L484

should read:
```js
      return new Array(x).fill(CLAIMING_ADDRESS_MIN_NANOWITS * 10 ** exp)
```

aesedepece (approved):

src/background.js
```diff
@@ -123,7 +123,7 @@ ipcMain.on('shutdown-finished', () => {
 
 // Exit cleanly on request from parent process in development mode.
 if (isDevelopment) {
-  main()
+  // main()
```
tmpolaczyk, line 126: ?

gabaldon (approved):

PR #1390, opened 2020-09-30
fix: update balances to new wallet API

mariocao:
## Why this change is necessary and useful

Update balances retrieval due to changes in the wallet API.


gabaldon:
duplicated in #1400

PR #1407, opened 2020-10-01
Fix: not show several addresses if only one repeated

### LINKED ISSUE #1395: Wrong fee when confirming send tx
When confirming a send tx the fee is not correct.
In this case the tx was
image

and the confirmation showed

image

### LINKED ISSUE #1397: Wrong several addresses tx
We found a case of a tx that is set as from several addresses but is actually from the same address, but different balances, and the different balances was not clear why.

image

clbartoli:
Some tx where set as several addresses but actually was just one repeated, now it shows only the address
Closes #1397

### DIFF
--- a/src/components/Transaction.vue
+++ b/src/components/Transaction.vue
@@ -167,7 +167,8 @@ export default {
   computed: {
     address() {
       if (this.type === 'POSITIVE') {
-        return this.inputs.length > 1
+        return this.inputs.length > 1 &&
+          this.inputs.some(input => input.address !== this.inputs[0].address)
           ? 'several addresses'
           : this.inputs[0].address
       } else {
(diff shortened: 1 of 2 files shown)


src/components/Transaction.vue
```diff
@@ -167,7 +167,8 @@ export default {
   computed: {
     address() {
       if (this.type === 'POSITIVE') {
-        return this.inputs.length > 1
+        return this.inputs.length > 1 &&
+          this.inputs.reduce((acc, input) => input.address !== acc.address)
```
tmpolaczyk, line 171: Here you want to check if there is at least one address different from the first one. You don't need reduce:

```suggestion
          this.inputs.some(input => input.address !== first_address)
```

src/components/Transaction.vue
```diff
@@ -167,7 +167,8 @@ export default {
   computed: {
     address() {
       if (this.type === 'POSITIVE') {
-        return this.inputs.length > 1
+        return this.inputs.length > 1 &&
+          this.inputs.some(input => input.address !== this.inputs[0])
```
tmpolaczyk, line 171:
```suggestion
          this.inputs.some(input => input.address !== this.inputs[0].address)
```

gabaldon (approved):

PR #1420, opened 2020-10-03
feat(claiming): update witnet email for each source

Tommytrg:


aesedepece (approved):

aesedepece:
Only suggestion is that we could take the chance to do something like this:
https://www.computerhope.com/issues/ch000056.htm

clbartoli (approved):

PR #1437, opened 2020-10-04
Fix electron single instance

gabaldon:


src/background.js
```diff
@@ -126,8 +126,28 @@ ipcMain.on('shutdown-finished', () => {
   app.quit()
 })
 
+const block = app.requestSingleInstanceLock()
+
+if (!block) {
+  app.quit()
+} else {
+  app.on('second-instance', (event, commandLine, workingDirectory) => {
+    // Si alguien intentó ejecutar un segunda instancia, debemos
```
clbartoli, line 135: mi no entender

mariocao:
On Linux it works like a charm!

(obviously, I had to give permissions to the `witnet` executable file)

Tommytrg (approved):

src/background.js
```diff
@@ -126,8 +126,25 @@ ipcMain.on('shutdown-finished', () => {
...
+const lock = app.requestSingleInstanceLock()
+
+if (!lock) {
+  app.quit()
+} else {
+  // if someone executes a second instace we should focus on the main window
+  app.on('second-instance', (event, commandLine, workingDirectory) => {
+    if (win) {
+      if (win.isMinimized()) win.restore()
+      win.focus()
```
mariocao, line 139: On Windows, it works that not a second instance is allowed. However, the `win.focus()` is not really working. It seems it tries to obtain the focus but it does not succeed (a sort of glitch).

PR #1446, opened 2020-10-06
 fix(claiming): use getUTCSeconds instead of getSeconds

Tommytrg:


src/utils.js
```diff
@@ -551,7 +551,7 @@ export function calculateVesting(vestingInfo, amount, genesisDate) {
     .map((_, index) => {
       const date = new Date(genesisDate)
       date.setSeconds(
```
tmpolaczyk, line 553:
```suggestion
      date.setUTCSeconds(
```

tests/unit/src/utils.spec.js
```diff
@@ -48,7 +48,7 @@ describe('calculateTimeAgo', () => {
   it('should end with "ago" copy', () => {
     const date = new Date()
 
-    date.setSeconds(date.getSeconds() - 4)
+    date.setSeconds(date.getUTCSeconds() - 4)
```
tmpolaczyk, line 51:
```suggestion
    date.setUTCSeconds(date.getUTCSeconds() - 4)
```

src/utils.js
```diff
@@ -550,8 +550,8 @@ export function calculateVesting(vestingInfo, amount, genesisDate) {
     .fill(0)
     .map((_, index) => {
       const date = new Date(genesisDate)
```
tmpolaczyk, line 552: `new Date` also uses local time, not UTC

src/utils.js
```diff
@@ -595,3 +596,14 @@ function factor(amount, base = 10, exp = 100) {
...
+export function convertDateToUtc (date = new Date()) {
+  return Date.UTC(
+    date.getUTCFullYear(),
+    date.getUTCMonth(),
+    date.getUTCDate(),
+    date.getUTCHours(),
+    date.getUTCMinutes(),
+    date.getUTCSeconds()
+  )
+}
```
Tommytrg, line 609:
```suggestion
```

src/utils.js
```diff
@@ -549,9 +549,10 @@ export function calculateVesting(vestingInfo, amount, genesisDate) {
   return Array(numberOfSteps)
     .fill(0)
     .map((_, index) => {
-      const date = new Date(genesisDate)
-      date.setSeconds(
-        date.getSeconds() + delay + cliff + installmentLength * index,
+      const date = convertDateToUtc(new Date(genesisDate))
```
Tommytrg, line 552:
```suggestion
      const date = new Date(genesisDate)
```

src/utils.js
aesedepece: I agree that using `Date` here offers no advantage. We can simply go with:
```js
const date = genesisDate + delay + cliff + installmentLength * index
```

aesedepece (changes requested):

tests/unit/src/utils.spec.js
```diff
@@ -49,7 +49,7 @@ describe('calculateTimeAgo', () => {
   it('should end with "ago" copy', () => {
     const date = new Date()
 
-    date.setSeconds(date.getSeconds() - 4)
+    date.setUTCSeconds(date.getUTCSeconds() - 4)
```
tmpolaczyk, line 52: This doesn't need to be changed, it's just a test that checks if dates from the past are displayed as "ago"

gabaldon (approved):

aesedepece (approved):

PR #1452, opened 2020-10-07
Delete auto-update

gabaldon:


gabaldon:
For the moment this feature is not going to be deleted. It will be reopened if needed.

PR #1456, opened 2020-10-07
Fix number format

Tommytrg:
image
image
image

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -113,7 +113,7 @@ export function standardizeWitUnits(
   amount,
   outputCurrency,
   inputCurrency = WIT_UNIT.NANO,
-  truncate,
+  truncate = 2,
 ) {
   // from input currency to output currency
   const witUnitConversor = {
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -135,9 +135,18 @@ export function standardizeWitUnits(
   }
   const num = Big(amount)
   const exponent = witUnitConversor[inputCurrency][outputCurrency]
-  return outputCurrency === WIT_UNIT.NANO
-    ? num.times(Big(10).pow(exponent), 0).toFixed(0)
-    : num.times(Big(10).pow(exponent), 0).toFixed(truncate)
+  const result = num.times(Big(10).pow(exponent), 0)
+
+  // console.log(result.toString(), result.cmp(1))
+  if (truncate === -1) Number(result.toString)
```
tmpolaczyk, line 141:
```suggestion
  if (truncate === -1) {
    return Number(result.toString)
  }
```

And personally I would use `null` as the special value instead of `-1`

gabaldon (dismissed):

tmpolaczyk (approved):

tmpolaczyk (approved):

PR #1480, opened 2020-10-21
Delete tray behavior in Sheikah

### LINKED ISSUE #1304: Discuss tray
We should discuss if it is necessary the tray behavior in Sheikah and if not remove it.

gabaldon:
close #1304


aesedepece:
This works really fine for me. It completely closes the app and the wallet backend.

I think this will save us from many headaches for now, until we come up with a more robust long-term solution for leaving Sheikah and the wallet running in the background.

:star: :star: :star: :star: :star:

clbartoli (approved):

clbartoli (approved):

PR #1485, opened 2020-10-26
Fix get tally transaction

### LINKED ISSUE #1484: Tallies returning funds have no `inputs`, leading to `undefined` error in `get_transactions`
**Describe the bug**
Tally transactions with no sufficient commit/reveals will return the funds to the data request creator and the amount will be `POSITIVE`. This will lead to an error while computing the transactions because there are no inputs in tally transactions and `inputs.length` is undefined.

The code in Sheikah for that behaviour is located in `function computeTransactionAddress(inputs, outputs, type)`. :stuck_out_tongue_winking_eye: 

**To Reproduce**
1. Create DR
2. DR has not sufficient commits
3. Sheikah will show an error notification and the transaction list will be empty

gabaldon:
close #1484

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -125,7 +125,6 @@ if (!lock) {
 
 // Exit cleanly on request from parent process in development mode.
 if (isDevelopment) {
-  main()
   if (process.platform === 'win32') {
     process.on('message', data => {
       if (data === 'graceful-exit') {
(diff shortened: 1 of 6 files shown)


src/background.js
```diff
@@ -145,7 +145,6 @@ if (!lock) {
 
 // Exit cleanly on request from parent process in development mode.
 if (isDevelopment) {
-  main()
```
clbartoli, line 148: is this no longer needed?

clbartoli (approved):

PR #1504, opened 2020-10-29
fix: waiting for node to sync status

### LINKED ISSUE #1496: Implement new wallet states (waiting for node to sync and sync error)
The new UI for showing the status of the wallet should also include the 'waiting for node to sync' status and the 'sync error.' It should show the re-sync functionality only in the latest case.

As @aesedepece suggested:

**Option 1**
Light (1)

gabaldon:
This pr adapt the new waiting for the node to sync status to the node status notification.
close #1496


gabaldon:
duplicated in #1529

PR #1505, opened 2020-10-30
Make amount input change currency on click and add milliWit

### LINKED ISSUE #1494: Sending a VTT
**Is your feature request related to a problem? Please describe.**
Sending a VTT right now is not very user friendly. I see a couple of problems:
1. An amount can be specified as e.g. 10 Wit, but if you then click on the unit to change to microWit or nanoWit, the amount does not scale. That is quite annoying if you want to specify an amount in Wit and a fee in e.g. microWit.
2. Using a `fee per weight unit` results in a pretty hard guess on how much the final fee will be. For example if I specify 1 microWit, will it be 1 microWit for the simplest VTT (1 input, 1 output) or 493 microWit(the weight for a 1 input / 1 output VTT). If it is a more complex transaction with multiple inputs and outputs this calculation becomes even harder.

**Describe the solution you'd like**
1. If a value has already been entered in either box, visually multiply or divide it when you click on the unit.
2. A couple of possible solutions here:
    1. Allow for specifying a flat fee
    2. In the confirmation box show what the final fee will be. Right now this seems to still be the `fee per weight` since it does not increase based on the inputs and outputs.

**Describe alternatives you've considered**
1. I guess it is possible to specify a fee as a floating point number but that is not too user friendly (if you want to denominate something in nanoWit for example).

PS: Why are we skipping milliWit, we are going from Wit to microWit to nanoWit? 😄

gabaldon:
close #1494

### DIFF
--- a/src/store/wallet.js
+++ b/src/store/wallet.js
@@ -51,6 +51,7 @@ export default {
     checkTokenGenerationEventDate: new Date(GENESIS_EVENT_TIMESTAMP),
     mainnetReady: false,
     currency: DEFAULT_WIT_UNIT,
+    prevCurrency: DEFAULT_WIT_UNIT,
     balance: {},
     walletIdx: null,
     sessionId: null,
@@ -136,7 +137,9 @@ export default {
       const unitsValues = Object.values(WIT_UNIT)
       const unitsKeys = Object.keys(WIT_UNIT)
       // Get index of the next currency
-      const index = (unitsValues.indexOf(state.currency) + 1) % 3
+      state.prevCurrency = state.currency
+      const index =
+        (unitsValues.indexOf(state.currency) + 1) % unitsValues.length
       state.currency = WIT_UNIT[unitsKeys[index]]
     },
     deleteSession(state) {
@@ -695,7 +698,6 @@ export default {
     },
     getWalletInfos: async function(context) {
       const request = await context.state.api.getWalletInfos()
-
       if (request.result) {
         context.commit('setWalletInfos', { walletInfos: request.result.infos })
       } else {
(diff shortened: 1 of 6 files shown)


src/store/wallet.js
```diff
@@ -166,7 +167,8 @@ export default {
       const unitsValues = Object.values(WIT_UNIT)
       const unitsKeys = Object.keys(WIT_UNIT)
       // Get index of the next currency
-      const index = (unitsValues.indexOf(state.currency) + 1) % 3
+      state.prevCurrency = state.currency
+      const index = (unitsValues.indexOf(state.currency) + 1) % 4
```
tmpolaczyk, line 171:
```suggestion
      const index = (unitsValues.indexOf(state.currency) + 1) % unitsValues.length
```

src/utils.js
```diff
@@ -127,16 +127,25 @@ export function standardizeWitUnits(
...
     },
+    [`${WIT_UNIT.MILLI}`]: {
+      [`${WIT_UNIT.WIT}`]: -3,
+      [`${WIT_UNIT.MILLI}`]: 0,
+      [`${WIT_UNIT.MICRO}`]: 3,
+      [`${WIT_UNIT.NANO}`]: 6,
+    },
     [`${WIT_UNIT.MICRO}`]: {
       [`${WIT_UNIT.WIT}`]: -6,
+      [`${WIT_UNIT.MILLI}`]: 6,
```
tmpolaczyk, line 142:
```suggestion
      [`${WIT_UNIT.MILLI}`]: -3,
```

Add tests

clbartoli (approved):

PR #1515, opened 2020-11-04
chore: allow set node url as env variable

Tommytrg:


tmpolaczyk:
The current approach will not work because the WITNET_NODE_URL environment variable is only used if the witnet.toml has the default node url. But, after using a custom node url, it is no longer the default one, so using the WITNET_NODE_URL again with a different url will not work.

The proposed solution is to remove the environment variable, and to only overwrite the node_url in witnet.toml one time, when the witnet.toml is first downloaded. Then, the user can simply modify the url in the witnet.toml file.

tmpolaczyk (approved):

PR #1535, opened 2020-11-20
Only close the wallet when Sheikah has run it

### LINKED ISSUE #1531: Sheikah should only tell the wallet backend to shutdown when spawned from Sheikah, never when running separately
Sheikah should only tell the wallet backend to shutdown when spawned from Sheikah, never when running separately

Tommytrg:
close #1531

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -427,7 +427,7 @@ async function runWallet() {
     walletProcess.stdout.on('data', async function(data) {
       console.info('stdout: ' + data.toString())
       status = STATUS.READY
-      win.webContents.send('loaded')
+      win.webContents.send('loaded', [{ isDefaultWallet: true }])
       await sleep(3000)
       loadUrl(status)
     })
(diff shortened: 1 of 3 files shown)


clbartoli (dismissed):

src/store/wallet.js
```diff
@@ -111,6 +112,9 @@ export default {
     },
   },
   mutations: {
+    setWalletOwner(status, { defaultWallet }) {
+      status.defaultWallet = defaultWallet
```
gabaldon, line 116:
```suggestion
      status.isDefaultWallet = defaultWallet
```

src/background.js
```diff
@@ -427,6 +426,7 @@ async function runWallet() {
     walletProcess.stdout.on('data', async function(data) {
       console.info('stdout: ' + data.toString())
       status = STATUS.READY
+      win.webContents.send('running')
```
gabaldon, line 429: Why are you sending this message right above loaded? Doing this you cannot differentiate the two steps.

gabaldon (changes requested):

gabaldon (dismissed):

PR #1541, opened 2020-11-24
feat: allow computing big numbers

gabaldon:
This pr is a workaround till issue [#1760](https://github.com/witnet/witnet-rust/issues/1760) is solved in witnet-rust


tests/unit/src/services/BigNumber.spec.js
```diff
@@ -0,0 +1,12 @@
+import BigNumber from '@/services/BigNumber'
+
+describe('should operate with big numbers', () => {
```
Tommytrg, line 3: My preference to test that is with a [scoped jest mock](https://jestjs.io/docs/en/manual-mocks#mocking-node-modules) of `big.js` library and check that the different operations of our wrapper call the correct method of `big.js`

Tommytrg (changes requested):

Tommytrg (approved):

PR #1542, opened 2020-11-25
fix: delete get transactions limit when there is a node movement

gabaldon:


src/store/wallet.js
```diff
@@ -751,15 +760,19 @@ export default {
...
+      const currentPage = Number.isInteger(payload.page)
+        ? payload.page
+        : context.state.currentTransactionsPage
+      const offset = (currentPage - 1) * 13
       const request = await context.state.api.getTransactions({
         wallet_id: context.state.walletId,
         session_id: context.state.sessionId,
-        limit,
-        page,
+        limit: 13,
```
Tommytrg, line 771: Could you move `13` to a constant?

tests/unit/src/components/TransactionList.spec.js
```diff
@@ -1,19 +1,129 @@
 import TransactionList from '@/components/TransactionList.vue'
 import Transaction from '@/components/Transaction.vue'
 // eslint-disable-next-line import/no-relative-parent-imports
-import transactionsMock from '../../mocks/transactions'
+
+const TransactionMock = {
```
Tommytrg, line 5: We already have mocked transactions, could you use or replace them with this?

Tommytrg (dismissed):

src/components/TransactionList.vue
```diff
@@ -76,6 +77,10 @@ export default {
       type: Array,
       required: true,
     },
+    totalTransactions: {
```
Tommytrg, line 80: Rename as transactionsLength

Tommytrg (approved):

PR #1566, opened 2020-12-10
feat: implement component to toggle notifications

gabaldon:


gabaldon:
included in #1578

src/store/wallet.js
```diff
@@ -476,14 +490,16 @@ export default {
...
-        createNotification({
-          title: 'Transaction succesfully sent',
-          body: `The transaction ${cropString(
-            transactionToSend.transaction_id,
-            12,
-            'middle',
-          )} has been sent succesfully into the Witnet network.\nIt should be written into a block soon.`,
-        })
+        if (context.state.notifications.transactions) {
+          createNotification({
```
tmpolaczyk, line 494: This API is a bad idea because we will forget to check if notifications are enabled when adding new notifications. It would be better to move this check inside `createNotification`, but since we want to support multiple "groups" of notifications, then we would need to pass one extra argument to `createNotification`.

So I would prefer an API like:

```suggestion
        createNotification({
            group: "transactions",
```

And in `createNotification` do something like:

```js
// Pseudocode
let group = notificationProps.remove(group);
if (!group) {
    // We forgot to pass a "group" parameter, what to do?
} else if (context.state.notifications.enabled[group]) {
    // Show notification
} else {
    // Notifications disabled, log to console?
}
```

PR #1567, opened 2020-12-10
fix: running wallet on cancel update

### LINKED ISSUE #1565: Stuck at 90% after cancelling update
I am using sheikah version b3142ffec8482536ab272fa5b110af7589baf486. That commit is version 1.5.0, and the latest version is 1.6.0. So, when opening sheikah, I see a message: "There is a new Sheikah version".

If I press "cancel" immediately, everything works as expected. But if I press "cancel" after the sheikah loading bar has reached 90%, it gets stuck at 90%.

gabaldon:
close #1565

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -417,7 +417,9 @@ function main() {
         win.webContents.send('downloaded')
         await sleep(3000)
       }
-      runWallet()
+      if (!isBeingUpdated) {
+        runWallet()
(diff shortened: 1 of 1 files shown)


tmpolaczyk (dismissed):
Works for me, thanks!

clbartoli (approved):

PR #1577, opened 2020-12-14
feat: allow send and receive strings when creating a transaction

### LINKED ISSUE #1575: Accept / send numbers as strings

gabaldon:
close #1575

### DIFF
--- a/src/components/CreateDataRequestForm.vue
+++ b/src/components/CreateDataRequestForm.vue
@@ -9,11 +9,7 @@
     width="max-content"
   >
     <el-form-item label="Witnesses" prop="witnesses">
-      <el-input
-        v-model="form.witnesses"
-        data-test="witnesses"
-        type="number"
-      ></el-input>
+      <el-input v-model="form.witnesses" data-test="witnesses"></el-input>
     </el-form-item>
 
     <el-form-item label="Collateral" prop="collateral">
@@ -29,7 +25,6 @@
       <el-input
         v-model="form.minConsensusPercentage"
         data-test="consensus"
-        type="number"
       ></el-input>
     </el-form-item>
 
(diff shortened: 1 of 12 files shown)


aesedepece:
This goes with https://github.com/witnet/witnet-rust/issues/1760

src/components/Send.vue
```diff
@@ -80,21 +81,24 @@ export default {
     FormInformation,
   },
   data() {
-    const integerNanoWit = (rule, value, callback) => {
-      const isNanoWit = this.currency === WIT_UNIT.NANO
-      if (isNanoWit && !Number.isInteger(Number(value))) {
-        callback(new Error('Only integer nanoWits values allowed'))
+    const maxNumber = (rule, value, callback) => {
+      const maxNumber = 18446744073709551615
```
tmpolaczyk, line 85: Since this number is greater than max safe integer, we must use a string to represent it. And since the next line does `if number >= maxNumber`, this maxNumber needs to be exactly `2**64`. Also, add a comment explaining what's this number and why is this the maximum.

```suggestion
      // 2**64. This is the max value of the u64 type in the wallet.
      const maxNumber = "18446744073709551616";
```

src/components/CreateDataRequestForm.vue
```diff
@@ -94,9 +86,12 @@ export default {
   },
   data() {
     const maxNumber = (rule, value, callback) => {
+      const maxNumber = 18446744073709551615
```
tmpolaczyk, line 89: This code is repeated twice, it should be extracted into a function like `isSmallerThan2ToThePowerOf64`, but with a better name

src/utils.js
```diff
@@ -143,15 +155,14 @@ export function standardizeWitUnits(
     const num = new BigNumber(amount)
     const exponent = witUnitConversor[inputCurrency][outputCurrency]
     const result = num.times(new BigNumber(10).pow(exponent), 0)
-
     if (truncate === -1) {
-      return result.toString()
+      return new BigNumber(result).toFixed()
```
tmpolaczyk, line 159:
```suggestion
      return result
```

tmpolaczyk: Wait, result is already a BigNumber? Then it should be 
```suggestion
      return result.toFixed()
```

aesedepece (approved):

tmpolaczyk (approved):

PR #1608, opened 2021-01-11
fix(TemplateList): show correct pagination

### LINKED ISSUE #1600: Fix DR templates pagination

Tommytrg:
- Fix TemplateList pagination

close #1600


clbartoli:
I think the genesis_block.json file is not needed in this pr

mariocao (changes requested):
I think the `genesis_block.json` shouldn't be there :smile:

src/components/Templates.vue
```diff
@@ -79,12 +81,19 @@ export default {
   },
   computed: {
     templatesPerPage() {
-      return this.currentPage === 1 ? 8 : 9
+      return 9
```
mariocao, line 84: Maybe a constant?

clbartoli:
I noticed that when creating a new template the name ("New template" by default) is not deleted when writing a different name. I'm not sure if it is on purpose, if not, may be it can be fixed in this PR, if not too complicated

image

Tommytrg:
@clbartoli When you create a new template, the editor assigns a default name: New template + N. Maybe we could discuss it on another issue if we want to change that field's current behavior. We could assign a default name but use it as a placeholder instead of as the current value.

gabaldon (approved):

PR #1612, opened 2021-01-12
feat: use nanoWit as static unit in fee field

### LINKED ISSUE #1597: Input fields for fees should always be expressed in priority value, and don't allow changing

Tommytrg:
- Lock unit change in fee field in Send.vue component
- Lock unit change in fee field in CreateDataRequestForm.vue component
- Use nanotWit as default currency

close #1597


gabaldon (commented):
The unit is changing when clicking to change unit in other fields. The amount should be also static.

gabaldon (approved):

PR #1614, opened 2021-01-13
Show pending confirmation status in transaction

### LINKED ISSUE #1593: Significant UX challenges regarding confirmation state of transactions
Current UX for confirmation state of transactions is quite challenging, almost impossible to reason about even for well-seasoned blockchain connoisseurs.

Namely, transactions that are included into block but not yet confirmed by superblocks appear indistinguishable from those that already confirmed.

Multiple users have already raised this issue, and in many cases it has lead to situations in which evil actors could potentially abuse the system to make some user believe they have received a payment they did not.

We need to urgently tackle this.

gabaldon:
Captura de pantalla 2021-01-13 a las 15 57 53

close #1593

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -283,12 +283,14 @@ export function standardizeTransactions(response) {
   // TODO(#1760): When the wallet is ready, it should receive the transaction amount, fees, inputs and outputs values as a string
   if (!response.result) return response
   const transactions = response.result.transactions.map(transaction => {
+    console.log('transaction', transaction)
     const transactionType = Object.keys(transaction.transaction.data)[0]
     const { inputs, outputs } = transaction.transaction.data[transactionType]
     // eslint-disable-next-line camelcase
     const { hash, miner_fee, block, timestamp, data } = transaction.transaction
     const tally = data[transactionType].tally
     const type = transaction.type
+    const confirmed = transaction.transaction.confirmed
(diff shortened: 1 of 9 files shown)


aesedepece:
Let me suggest a tooltip with the following text: _"This transaction is now in the block chain. However, for it to be considered final and irreversible, a majority of the network needs to confirm it. This normally takes no more than 15 minutes."_

By the way, what's the UI showing for transactions that have been sent recently and are not in any block just yet? I think we could also use the yellow label, but change it to _"Sending"_, with a tooltip saying _"This transaction is waiting to be included in the block chain. Depending on the chosen priority and how clogged is the network, this may take from a few seconds to several hours."_ 

In the future we can integrate that with some API that will give us an ETA on time-to-chain of a transaction based on its priority; as well as enabling users to do _replace-by-fee_ if they messed up the priority.

src/api.js
```diff
@@ -283,12 +283,14 @@ export function standardizeTransactions(response) {
   // TODO(#1760): When the wallet is ready, it should receive the transaction amount, fees, inputs and outputs values as a string
   if (!response.result) return response
   const transactions = response.result.transactions.map(transaction => {
+    console.log('transaction', transaction)
```
Tommytrg, line 286: Remove console.log

Tommytrg (approved):

PR #1615, opened 2021-01-14
fix: templates list layout

gabaldon:


Tommytrg (changes requested):
- Add a commit for fix Transactions layout
- When I remove the last template the button disappears

Tommytrg (approved):

PR #1622, opened 2021-01-15
ref(Address): implement Address component to unify logic and style

Tommytrg:
- Implement an Address component and refactor to avoid duplication

### DIFF
--- a/src/components/Address.vue
+++ b/src/components/Address.vue
@@ -1,0 +1,42 @@
+<template>
+  <p class="address" :style="{ fontSize: size }" data-test="address">
+    {{ address }}
+  </p>
+</template>
+
+<script>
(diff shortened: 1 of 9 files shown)


src/components/Address.vue
```diff
@@ -0,0 +1,43 @@
...
+    size: {
+      type: String,
+      default: '12px',
+    },
+  },
+  computed: {
+    address() {
+      return this.blind ? this.blindAddress : this.value
+    },
+    blindAddress() {
```
gabaldon, line 28: you can use ```cropString``` from utils to do so

clbartoli:
When open transaction details the view is displaced 
image

gabaldon (approved):

PR #1628, opened 2021-01-20
feat(AddressInformation): extend time for "Copied!" feedback

### LINKED ISSUE #1580: "Copied" transaction tooltip disappearing too fast
Steps to reproduce:

1. Generate new address.
2. Hover the mouse over the copy button. It shows a tooltip that says "Copy" until you move the mouse somewhere else.
3. Click on the copy button. It shows a tooltip that says "Copied", but it disappears after about 0.5 seconds, so it's easy to miss.

Ideally, the tooltip should be visible, or sheikah should show a different indicator. For example, a message like "address copied to clipboard" on the bottom of the screen.

aesedepece:
Fix #1580

### DIFF
--- a/src/components/AddressInformation.vue
+++ b/src/components/AddressInformation.vue
@@ -159,7 +159,7 @@ export default {
       this.copied = true
       setTimeout(() => {
         this.copied = false
-      }, 2000)
+      }, 5000)
     },
     formatDateVerbose,
   },


tmpolaczyk:
This doesn't fix the timeout of the tooltip, it only fixes the timeout of the green checkmark that replaces the copy button after pressing copy. It's a strange issue, because moving the mouse outside of the checkmark and then back again inside results in a tooltip that does not disappear.

Additionally, the checkmark does not disappear when selecting a different address, and you have to wait a bit until it lets you press the copy button to copy the different address.

Tommytrg:
I think this could be fixed by passing the prop `hide-after` to the `el-tooltip` component in AddressInformation.vue.

https://element.eleme.io/#/en-US/component/tooltip#attributes

gabaldon:
included in #1650

PR #1629, opened 2021-01-20
fix(locales): add html interpolation for conflicting strings

gabaldon:


.eslintrc.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     'vue/dot-location': ['error', 'property'],
     'vue/eqeqeq': 'error',
     'vue/key-spacing': 'error',
+    'vue/no-v-html': 'off',
```
tmpolaczyk, line 31: Using `v-html` allows XSS attacks so disabling this lint may be dangerous.

src/components/BalanceData.vue
```diff
@@ -12,9 +12,9 @@
       <el-tooltip trigger="hover" effect="light">
         <font-awesome-icon class="info" icon="info-circle" />
         <div slot="content" class="info-message">
-          {{ this.$t('locked_balance_tooltip[0]') }}
+          {{ this.$t('locked_balance_tooltip_0') }}
           <a class="link" :href="balanceLockedUrl" target="_blank"
-            >{{ this.$t('locked_balance_tooltip[1]') }}
+            >{{ this.$t('locked_balance_tooltip_1') }}
```
aesedepece, line 17: This is still conflicting, as it assumes that the link saying "here" goes at the end in every language.

Same for all the rest. The point of HTML interpolation is to write the HTML tags in the strings found in the `en.json` file, so translators can freely decide how they must be used in their language.

aesedepece (changes requested):

src/components/BalanceData.vue
aesedepece: Wait, wait wait, forget about that. I'm reading the [component interpolation guide](https://kazupon.github.io/vue-i18n/guide/interpolation.html#basic-usage) now. Let me see whether this will work or not.

gabaldon:
duplicated in #1649

PR #1632, opened 2021-01-21
Update witnet-radon-js and simplify SettingsLanguage

Tommytrg:
This PR have to be merged after [witnet#witnet-radon-js/44](https://github.com/witnet/witnet-radon-js/pull/44) and witnet-radon-js v0.8.2 is released.

- Upgrade `witnet-radon-js` to 0.8.4. Now Radon class accepts a locale argument to set the language of the descriptions. Also, expose a method `setLocale`, which we call to update the language of the current template.
- Simplify language settings flow. Now we store the locale, and we derive the language from it instead of doing it the other way around. This change makes it easier to add specific country locales in the future.

### DIFF
--- a/src/store/wallet.js
+++ b/src/store/wallet.js
@@ -22,7 +22,7 @@ import {
   NETWORK_STATUS,
   TRANSACTIONS_LIMIT,
   NOTIFICATIONS,
-  DEFAULT_LANGUAGE,
+  DEFAULT_LOCALE,
   LANGUAGES,
   LOCALES,
   THEMES,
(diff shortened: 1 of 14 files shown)


src/store/wallet.js
```diff
@@ -873,16 +879,12 @@ export default {
...
-          i18n: payload.i18n,
-        })
-      } else {
-        context.commit('setLanguage', {
-          language: null,
+    getLocale: async function(context, payload) {
+      const locale = context.state.localStorage.getLanguageSettings()
+
+      if (locale) {
+        context.commit('setLocale', {
```
gabaldon, line 886:
```suggestion
        context.commit('setLanguage', {
```

gabaldon (approved):

PR #1646, opened 2021-02-03
fix(router): do not redirect to wallet list if session id

### LINKED ISSUE #1644: Sheikah goes back to the wallet list view unexpectedly
**Describe the bug**
Sometimes Sheikah goes back to the "Wallet list" view in order to request the user to unlock a wallet.
This happens even if the user has an already unlocked wallet, thus triggering potential "double synchronization processes".

**To Reproduce**
This behaviour seems a little bit random, but we managed to reproduced it several times as follows:
1. Press "Resync" button
2. Restart the wallet process (while resyncing)
3. Sheikah detects the disconnection and asks the user to unlock the wallet again
4. As soon as that happens, unlock the wallet (it has to be pretty fast)
5. Sheikah shows the transactions view but it kicks the user back to the "Wallet list" view

As side note we also experienced the same behaviour while playing with the resync button and the transactions view.
We weren't able to find a way to reproduce it all the times. Maybe it had something to to do with the fact that we stopped the wallet several times.

gabaldon:
close #1644

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -100,12 +100,15 @@ export default new Router({
             }
           }, 3000)
           store.state.wallet.api.client.ws.on('open', async () => {
-            error = false
-            await store.dispatch('getWalletInfos')
             const polling = setInterval(async () => {
+              error = false
               clearInterval(polling)
+              await store.dispatch('getWalletInfos')
               const walletInfos = store.state.wallet.walletInfos
-              if (walletInfos && walletInfos.length > 0) {
+              const isSessionId = store.state.wallet.sessionId
+              if (isSessionId) {
+                next()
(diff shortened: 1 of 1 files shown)


tmpolaczyk (dismissed):
Tested, seems to be working fine

tmpolaczyk (changes requested):
After the latest changes, it doesn't fix the problem.

Steps to reproduce:

* Start sheikah
* Wait for "wallet disconnected screen"
* Start wallet
* Enter password fast using ctrl-v and enter
* After about 1 second, sheikah goes back to the main screen and asks for the password again

tmpolaczyk (approved):

PR #1650, opened 2021-02-19
Fix copied tooltip delay

### LINKED ISSUE #1580: "Copied" transaction tooltip disappearing too fast

gabaldon:
Fix #1580

### DIFF
--- a/src/components/AddressInformation.vue
+++ b/src/components/AddressInformation.vue
@@ -20,6 +20,8 @@
       </el-tooltip>
       <el-tooltip
         v-if="!used && copied"
+        v-model="copied"
+        manual
         :content="$t('copied')"
         placement="bottom"
         effect="light"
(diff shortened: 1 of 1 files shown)


tmpolaczyk (approved):
Works!

Only a few minor notes:
* There is no mention of the word "clipboard", so some users may wonder where it was copied to.
* After pressing copy you can change to another address, and the other address cannot be copied until the 3 second timeout expires.

PR #1662, opened 2021-04-05
fix(locales): delete unnecessary constants

gabaldon:


src/components/BalanceButtons.vue
```diff
@@ -37,7 +36,7 @@ export default {
        * Emitted when send button is clicked
        * @event send
        */
-      if (this.currentStatus === NETWORK_STATUS.SYNCED) {
+      if (this.currentStatus === this.$t('synced')) {
```
Tommytrg, line 39: We should only use translations in the view. If we don't have a translation for currentStatus the sentence could fail in the future.

Tommytrg (approved):

PR #1674, opened 2021-04-26
Order transactions by timestamp and time-lock

### LINKED ISSUE #1672: Order transactions by timelock
We should paginate transactions from the wallet to allow ordering transactions by timestamp and timelock.

gabaldon:
close #1672


aesedepece (dismissed):
`Tested ACK`

PR #1686, opened 2021-05-17
fix: show connected to node address when it is connected

### LINKED ISSUE #1685: "Trying to connect" when already connected
See in this comment:
https://github.com/witnet/sheikah/pull/1680#pullrequestreview-660736791

"Trying to connect" is shown during synchronization.

I was expecting to see "Connected" as soon as the connection between the wallet backend and the node is established.

gabaldon:
close #1685

### DIFF
--- a/src/components/NetworkStatus.vue
+++ b/src/components/NetworkStatus.vue
@@ -83,7 +83,7 @@
           {{ this.$t('of') }}
           <span class="bold">#{{ lastBlock }}</span>
         </p>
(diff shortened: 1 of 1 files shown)


aesedepece (commented):
Will try this now.

aesedepece:
Having an address is not enough. We may need to check for the node connection status. See how it says it's connected and disconnected at the same time:
image

aesedepece (approved):
Sweet!

PR #1693, opened 2021-05-24
feat: allow sending vtt with timelock

### LINKED ISSUE #1692: Add timelock support while sending funds (value transfer transactions)
While creating a new Value Transfer Transaction (i.e. using "Send"), users should be able to optionally define a date for the time lock field (i.e. date in which the send amount will be unlocked).

Some technical details:
- The date should be selected with a Date Picker component
- If not defined, timelock field is not required by the Wallet endpoint (is optional)

gabaldon:
close #1692

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -96,8 +96,7 @@ export class WalletApi {
   }
 
   createVTT(params) {
-    const defaultParams = { time_lock: 0 }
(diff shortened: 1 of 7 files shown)


mariocao:
It works!!! 

As discussed, we could show the detected timezone instead of the UTC. :)

mariocao (approved):
:fireworks:

PR #1694, opened 2021-05-26
feat: check if the wallet version to download is compatible

gabaldon:


src/background.js
```diff
@@ -390,12 +391,20 @@ function main() {
...
           const versionFile = fs.readFileSync(
             path.join(SHEIKAH_PATH, VERSION_FILE_NAME),
             'utf8',
           )
+          const diff = semver.diff(
+            getVersionFromName(versionFile),
+            getVersionFromName(latestReleaseVersion),
+          )
+          isLatestVersionCompatible = !diff || diff === 'patch'
+
```
Tommytrg, line 407:
```suggestion
          isLatestVersionCompatible = semver.satisfies(getVersionFromName(latestReleaseVersion), `~${getVersionFromName(versionFile)}`)

```

tmpolaczyk: Using ~1.2.1 will make version 1.2.1 compatible with 1.2.2, but not with 1.3.0 and not with 2.0.0. Is that what we want?

gabaldon: I think so

tmpolaczyk (dismissed):

tmpolaczyk (approved):

PR #1695, opened 2021-05-26
fix: show password validation error as a message not an alert

gabaldon:


tmpolaczyk:
If "Validation Error" does not need to be translated, maybe we should remove it from the translation files?

Tommytrg (approved):

PR #1697, opened 2021-05-27
feat: clear password input when clicking export xprv

gabaldon:


Tommytrg (changes requested):
AFAIK, we don't need to pass a prop to call the validate method. You directly call a component instance method from the parent function instead.

src/components/ExportXprv.vue
```diff
@@ -58,6 +59,7 @@ export default {
   data() {
     return {
       password: '',
+      clearInput: false,
```
Tommytrg, line 62: We don't need this anymore

src/components/ExportXprv.vue
```diff
@@ -124,6 +127,7 @@ export default {
       this.clear()
     },
     async encryptAndExport() {
+      // FIXME: clear password
```
Tommytrg, line 130: I don't understand what this means

Tommytrg (approved):

PR #1702, opened 2021-06-08
fix: improve timelock ux

gabaldon:
This pr implements the suggestions from discussion #1701. 
The element-UI locale updates itself when the language changes and the date-picker input is hidden now behind the advance options button.

Captura de pantalla 2021-06-08 a las 15 48 18
Captura de pantalla 2021-06-08 a las 15 48 40

### DIFF
--- a/src/components/SendValueTransferForm.vue
+++ b/src/components/SendValueTransferForm.vue
@@ -75,7 +75,7 @@
       </el-button>
       <el-button
         class="send-btn"
-        tabindex="5"
+        tabindex="6"
         type="primary"
         data-test="sign-send-btn"
         @click="tryCreateVTT"
@@ -314,6 +314,7 @@ export default {
   .link {
     font-size: 14px;
     text-align: left;
+    width: fit-content;
 
     .icon {
       width: 8px;
--- a/src/components/SendValueTransferForm.vue
+++ b/src/components/SendValueTransferForm.vue
@@ -35,19 +35,46 @@
         <AppendUnit slot="append" :static-unit="WIT_UNIT.NANO" />
       </el-input>
     </el-form-item>
-    <el-form-item :label="$t('timelock')" prop="timelock">
-      <el-date-picker
-        v-model="form.timelock"
-        type="datetime"
-        placeholder="Select date and time"
-        tabindex="5"
-        :default-value="new Date()"
-        value-format="timestamp"
-      />
-    </el-form-item>
+    <transition name="slide">
+      <el-form-item
+        v-if="isAdvancedVisible"
+        :label="$t('timelock')"
+        prop="timelock"
+      >
+        <el-date-picker
+          v-model="form.timelock"
+          type="datetime"
+          :placeholder="$t('select_date')"
+          tabindex="5"
+          :default-value="new Date()"
+          value-format="timestamp"
+        />
+      </el-form-item>
+    </transition>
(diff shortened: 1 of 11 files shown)


aesedepece:
Can we use `▲` and `▼`  or `▴` and `▾` on the "Advanced options" label? I think it helps understanding that it can be toggled on / off.

aesedepece:
The "Now" option inside the date picker seems redundant to me.

If we can't remove it, I'd rename it to "No timelock" (of course, using the current date as timelock will equal no timelock in practice).

As the timelock is an optional feature and has two very clearly differentiated valid values (no date vs. some date), I'd also appreciate a quick way to clear this field. Something like an `x` inside the text box or next to it would work for me.

clbartoli (dismissed):

src/components/SendValueTransferForm.vue
```diff
@@ -35,19 +35,46 @@
...
+        class="link"
+        type="text"
+        @click="toggleAdvanceOptions"
+      >
+        {{ this.$t('show_advance') }}
+        <CustomIcon class-name="icon" name="open" />
+      </el-button>
+      <el-button
+        class="send-btn"
         tabindex="5"
```
Tommytrg, line 78:
```suggestion
        tabindex="6"
```

src/store/wallet.js
```diff
@@ -28,6 +28,7 @@ import {
 } from '@/constants'
 import { SET_TEMPLATES, UPDATE_TEMPLATE } from '@/store/mutation-types'
 import warning from '@/resources/svg/warning.png'
+import locale from 'element-ui/lib/locale'
```
Tommytrg, line 31:
```suggestion
import elementLocale from 'element-ui/lib/locale'
```

src/store/wallet.js
```diff
@@ -247,15 +248,17 @@ export default {
...
+        state.locale = newLocale
+        state.localStorage.setLanguageSettings(newLocale)
 
-        i18n.locale = locale
+        i18n.locale = newLocale
       } else {
         console.warn('[mutation setUnit]: invalid language')
       }
+      // Set element locale
+      locale.use(LANGUAGES[i18n.locale].elementLocale)
```
Tommytrg, line 261:
```suggestion
      elementLocale.use(LANGUAGES[i18n.locale].elementLocale)
```

src/styles/element-variables.scss
```diff
@@ -358,6 +358,10 @@ $--font-path: '~element-ui/lib/theme-chalk/fonts';
   background: var(--date-picker-background);
   border-top: var(--date-picker-border);
   color: var(--text-medium-emphasis);
+
+  .el-button--text {
+    display: none;
```
Tommytrg, line 363: Why do we need this rule?

src/components/SendValueTransferForm.vue
```diff
@@ -35,20 +35,47 @@
...
+          :default-value="new Date()"
+          value-format="timestamp"
+        />
+      </el-form-item>
+    </transition>
     <p v-if="createVTTError" class="error">{{ createVTTError.message }}</p>
     <div class="submit">
       <el-button
-        tabindex="5"
+        v-if="isAdvancedVisible"
```
Tommytrg, line 57: Add css rule `width: fit-content` to avoid wrong behavior clicking outside of the button

clbartoli (approved):

PR #1705, opened 2021-06-28
feat: allow to set supported witnet-rust version in code

tmpolaczyk:

It allows new versions of sheikah to update to new versions of witnet-rust


src/background.js
```diff
@@ -400,9 +400,11 @@ function main() {
             'utf8',
           )
 
+          const supportedVersion = "1.2"
```
gabaldon, line 403: Create this constant in line 45
```suggestion
          const SUPPORTED_VERSION = '1.2'
```

tmpolaczyk: Called it `WITNET_RUST_VERSION` to be more explicit.

gabaldon (approved):

PR #1706, opened 2021-06-29
feat: add data request examples

gabaldon:


src/store/rad.js
```diff
@@ -468,6 +484,7 @@ export default {
           key: 'templates',
           value: templates,
         })
+        console.log(request)
```
clbartoli, line 487: is this console.log needed?

src/constants.js
```diff
@@ -163,3 +166,21 @@ export const BIRTH_DATE_DELAY_DAYS = 30
...
+    description: 'Bitcoin price in USD',
+    radRequest: bitcoinPrice,
+  },
+  {
+    name: 'Eth price',
+    description: 'Eth price in USD',
+    radRequest: ethPrice,
+  },
+  {
+    name: 'Gold price',
```
clbartoli, line 182: I would delete the gold example, I think the BTC and ETH are enough

src/components/EditorToolBar.vue
```diff
@@ -156,14 +156,17 @@ export default {
...
-      set() {
-        this.toggleTryDataRequest()
-      },
       get() {
         return this.autoTry
       },
+      set() {
+        this.toggleTryDataRequest()
+      },
     },
```
Tommytrg, line 165: AFAIK we are not changing anything here

src/components/steps/Loading.vue
```diff
@@ -30,6 +31,7 @@ export default {
     sessionId(value) {
       if (value) {
         this.$router.push(`/wallet/transactions?session_id=${this.sessionId}`)
+        this.setDefaultTemplates({ locale: this.locale })
```
Tommytrg, line 34: IMO, this logic doesn't belong to the loading component. We can call this method after unlock the wallet in vuex. There we should also check if the wallet already has the default templates before save them.

src/store/rad.js
```diff
@@ -489,6 +505,10 @@ export default {
         session_id: context.rootState.wallet.sessionId,
         key: 'templates',
       })
+      console.log(
+        `getTemplates--${context.rootState.wallet.walletId}-${context.rootState.wallet.sessionId}--`,
+        request,
+      )
```
Tommytrg, line 511:
```suggestion
```

Tommytrg (changes requested):

src/components/steps/Loading.vue
gabaldon: #1714

Tommytrg (approved):

Tommytrg:
Merged in #1715

PR #1720, opened 2021-09-24
fix: ftu download only wallet compatible version

gabaldon:


src/background.js
```diff
@@ -62,6 +62,8 @@ const WITNET_FILE_NAME = {
 const WITNET_CONFIG_FILE_NAME = 'witnet.toml'
 const LATEST_RELEASES_URL =
   'https://api.github.com/repos/witnet/witnet-rust/releases/latest'
+const LATEST_RELEASE_COMPATIBLE =
```
Tommytrg, line 65: This is not the lastest release compatible, if witnet-rust publishes a patch release it will be the latest
```suggestion
const COMPATIBLE_RELEASE =
```

src/background.js
```diff
@@ -354,7 +356,14 @@ function loadUrl(status) {
 
 function main() {
   console.info('Fetching releases from: ' + LATEST_RELEASES_URL)
-  axios.get(LATEST_RELEASES_URL).then(async result => {
+  const existVersionFile = fs.existsSync(
```
Tommytrg, line 359: Remove redeclaration of `existVersionFile` [line 393](https://github.com/witnet/sheikah/pull/1720/files#diff-16f022ec89282b37542e3374910a5e4a4cf37596abf8f723f0d7f8ab1096f853R393)

Tommytrg (approved):

PR #1736, opened 2021-11-30
Add support for rng operator

gabaldon:


Tommytrg:
IMO we should remove the add source button when the source type is RNG because currently, we can't have an RNG request with multiple sources.

Tommytrg (approved):

PR #1749, opened 2021-12-20
fix(background): disable chrome shortcuts on production

Tommytrg:


src/background/appManager.ts
```diff
@@ -170,7 +170,7 @@ export class AppManager {
 
     this.win.on('close', this.closeApp.bind(this))
 
-    if (DEVELOPMENT) {
+    if (!DEVELOPMENT) {
```
tmpolaczyk, line 173:
```suggestion
```

Why not show it always?

gabaldon (approved):

src/background/appManager.ts
Tommytrg: We decided to disable chrome shortcuts to avoid users opening the console or zooming in in order to give a native app feeling

tmpolaczyk: This code does not disable shortcuts, it enables the menu.

Tommytrg: It desables the shortcuts creating a new menu which overloads them with empty functions

tmpolaczyk: A comment about that would be nice.

PR #1774, opened 2022-03-29
feat: bump witnet-rust version to 1.5

gabaldon:


aesedepece (approved):
utACK

src/background/walletManager.ts
```diff
@@ -53,30 +53,31 @@ export class WalletManager {
...
+  public async run() {
     if (this.existDirectory) {
+      // Check if latest version is compatible or needs to be downloaded
       try {
         const versionName = fs.readFileSync(
           path.join(SHEIKAH_PATH, VERSION_FILE_NAME),
           'utf8',
         )
         const installedVersion = getVersionFromName(versionName)
+        const latestWitnetRustVersion = await getLatestWitnetRustRelease()
```
Tommytrg, line 68: What will happen if we receive a timeout from Github? We should handle that error and don't update the wallet in this case

Tommytrg: > What will happen if we receive a timeout from Github? We should handle that error and don't update the wallet in this case

This function is inside a try-catch

PR #1798, opened 2022-08-29
fix(WalletList): avoid changing selected wallet on get walletInfos

### LINKED ISSUE #1790: Login screen will periodically change the wallet in the select box to #1
**Describe the bug**
While in the login screen, if you have multiple wallets, some backend call is causing the select box to re-render and default back to the wallet #1 being selected.

This is particularly annoying when it happens while typing the password, as the selection change will most probably go unnoticed, and then you'll get an "invalid password" error which can be quite confusing.

**To Reproduce**
1. Go to the Sheikah login view
2. Make sure that you have multiple wallets
3. Change the selected wallet
4. Wait and see

**Expected behavior**
The user selection should be honored.

**Additional context**
This bug has been known since the testnet times, but it was probably never fixed because we never filed a proper issue for it.

gabaldon:
close #1790

### DIFF
--- a/src/components/WelcomeBack/WalletList.vue
+++ b/src/components/WelcomeBack/WalletList.vue
@@ -92,12 +92,6 @@ export default {
     },
   },
   watch: {
-    wallets: {
-      deep: true,
-      handler() {
-        this.currentWallet = this.walletOptions[this.lastWalletOpen]
-      },
-    },
     sessionId: function (newValue) {
       if (newValue) {
         this.updateView()


src/components/WelcomeBack/WalletList.vue
```diff
@@ -94,8 +94,10 @@ export default {
   watch: {
     wallets: {
       deep: true,
-      handler() {
-        this.currentWallet = this.walletOptions[this.lastWalletOpen]
+      handler(prevValue, newValue) {
+        if (prevValue.length !== newValue.length) {
+          this.currentWallet = this.walletOptions[this.lastWalletOpen]
+        }
```
aesedepece, line 100: Would it be possible to actually resolve `prevValue[this.lastWalletOpen]`, catch some wallet identifier and try to find its index in `newValue`? Or would it be too overkill?

aesedepece (approved):
This did the trick for me!

aesedepece (approved):

src/components/WelcomeBack/WalletList.vue
gabaldon: Actually, I don't think the watcher is necessary at all, I will remove it!

aesedepece (approved):

PR #1801, opened 2022-09-06
feat(estimatedFee): allow selecting estimated fees

### LINKED ISSUE #1799: Add miner fee estimation to vtt and dr

gabaldon:
- [x] Allow selecting estimated fee in vtt
- [x] Allow selecting estimated fee in dr

Screenshot 2022-09-06 at 18 52 47

close #1799


src/components/SelectEstimatedFee.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <div v-if="estimationOptions" class="estimation-table">
```
Tommytrg, line 2: Could we move this `if` condition to the parent ([SetValueTransferFee.vue](https://github.com/witnet/sheikah/pull/1801/files#diff-0b5cb4cab008d30d8674a21ff613be33644240d94a8aee525473ea863502beb4))? If we do that, we will render this component only when it's necessary

aesedepece:
The ]PR on witnet-rust](https://github.com/witnet/witnet-rust/pull/2261) has been revamped with a new approach. It know has fixed time-to-block intervals (1m, 5m, 15m, 1h, 6h), and they are all expressed in seconds.

tmpolaczyk:
Is this how it should look like in dark mode? I expected some border around the other options.

Captura de pantalla de 2022-09-27 14-31-15

aesedepece:
> I bumped into this error open_mouth image

Oh I think it's because my locally compiled Sheikah believes it's on testnet but it's actually on mainnet, hence the address length mismatch.

src/components/SetFee.vue
```diff
@@ -0,0 +1,389 @@
...
+        return {
+          label: trait,
+          result: await transaction,
+        }
+      })
+      Promise.all(txRequests).then(result => {
+        this.estimatedTransactions = result.reduce((acc, tx) => {
+          acc[tx.label] = tx.result
+          return acc
+        }, {})
```
aesedepece, line 254: Just in case you're a fan of one-liners:
```suggestion
        this.estimatedTransactions = result.reduce((acc, tx) => ({...acc, [tx.label]: tx.result}), {})
```

src/locales/en.json
```diff
@@ -420,5 +420,13 @@
   "error_validating_delete": "The typed name does not match the name of the current wallet",
   "fee_info": "By default, 'Weighted fee' is selected. The amount of the fee will be calculated, taking into account the weight of the transaction. To set an absolute fee, you need to toggle 'Absolute fee' in the advance options below.",
   "weighted_fee": "Fee per weight unit",
-  "absolute_fee": "Absolute fee"
+  "absolute_fee": "Absolute fee",
+  "set_vtt_miner_fee": "Select Value Transfer Transaction fee",
+  "set_dr_miner_fee": "Select Data Request Transaction fee",
```
aesedepece, line 425:
```suggestion
  "set_vtt_miner_fee": "Choose your desired miner fee",
  "set_dr_miner_fee": "Choose your desired miner fee",
```

src/locales/en.json
```diff
@@ -420,5 +420,13 @@
...
+  "get_estimation_error_message": "Error estimating the miner fees",
```
aesedepece, line 426:
```suggestion
  "get_estimation_error_message": "Error while estimating the miner fee",
```

src/locales/es.json
```diff
@@ -418,5 +418,12 @@
   "error_validating_delete": "El nombre escrito no coincide con el nombre de esta wallet",
   "fee_info": "Por defecto, la tasa se calcula teniendo el cuenta el peso de la transacción. Para poder enviar una transacción con tasa absoluta selecciona 'Tasa absoluta' en el switch que aparece debajo.",
   "weighted_fee": "Tasa por unidad de peso",
-  "absolute_fee": "Tasa absoluta"
+  "absolute_fee": "Tasa absoluta",
+  "set_vtt_miner_fee": "Selecciona la tasa para la transacción",
```
aesedepece, line 422:
```suggestion
  "set_vtt_miner_fee": "Elige la tasa de minado",
```

aesedepece (changes requested):

aesedepece:
> > I bumped into this error open_mouth image
> 
> Oh I think it's because my locally compiled Sheikah believes it's on testnet but it's actually on mainnet, hence the address length mismatch.

No, it perfectly knows it's on mainnet, because it will complain exactly about that if I provide a testnet address :thinking:

tmpolaczyk:
> Oh I think it's because my locally compiled Sheikah believes it's on testnet but it's actually on mainnet, hence the address length mismatch.

I don't think it's a compile time option, I sometimes use the release build in testnet and it works fine.

aesedepece:
> > Oh I think it's because my locally compiled Sheikah believes it's on testnet but it's actually on mainnet, hence the address length mismatch.
> 
> I don't think it's a compile time option, I sometimes use the release build in testnet and it works fine.

Looks like I was wrong twice: it's like my wallet knows it's on mainnet, but the frontend doesn't:
image

aesedepece:
> > > Oh I think it's because my locally compiled Sheikah believes it's on testnet but it's actually on mainnet, hence the address length mismatch.
> > 
> > 
> > I don't think it's a compile time option, I sometimes use the release build in testnet and it works fine.
> 
> Looks like I was wrong twice: it's like my wallet knows it's on mainnet, but the frontend doesn't: image
> 
> In any case, doesn't seem related to this issue, so I won't bother more and figure out a solution on my own sweat_smile

Just for the record, and sorry for spamming: running `./dist_electron/Sheikah\ Witnet\ wallet-1.11.3.AppImage` instead of `yarn electron:serve` or `./dist_electron/linux-unpacked/sheikah-witnet-wallet ` seems to do the trick.

src/components/SetFee.vue
```diff
@@ -0,0 +1,389 @@
...
+    estimationOptions() {
+      if (this.estimatedTransactions && this.formatedFeeEstimationReport) {
+        const result = FEE_TRAITS.reduce((acc, trait) => {
+          acc.push({
+            label: trait,
+            report: this.formatedFeeEstimationReport[trait],
+            transaction: this.estimatedTransactions
+              ? this.estimatedTransactions[trait]
+              : {},
+          })
```
aesedepece, line 179:
```suggestion
        const result = FEE_TRAITS.reduce((acc, tier) => {
          acc.push({
            label: tier,
            report: this.formatedFeeEstimationReport[tier],
            transaction: this.estimatedTransactions
              ? this.estimatedTransactions[tier]
              : {},
          })
```

src/constants.js
```diff
@@ -14,6 +14,8 @@ export const EDITOR_EXPORT_FORMAT = {
 export const TRANSACTIONS_LIMIT = 13
 export const EDITOR_TRY_INTERVAL = 3000
 export const EDITOR_SAVE_INTERVAL = 3000
+export const SECONDS_TO_NEXT_BLOCK = 45
+export const FEE_TRAITS = ['stinky', 'low', 'medium', 'high', 'opulent']
```
aesedepece, line 18:
```suggestion
export const FEE_TIERS = ['stinky', 'low', 'medium', 'high', 'opulent']
```

aesedepece (changes requested):

aesedepece:
Let's take into account that the estimates report that comes back from the `priority` method only speaks about priority. These cannot be shown as fees without first multiplying them by the transaction weight.

Hence this functionality cannot be implemented without further changes in witnet-rust, so as to return the transaction weight upon calling `send_transaction`.

tmpolaczyk:
> Let's take into account that the estimates report that comes back from the priority method only speak about priority. These cannot be shown as fees without first multiplying them by the transaction weight.

Agree that it would be nice to show the final cost, but sheikah already supports relative fee ("Fee per weight unit"), which is the same concept as priority.

aesedepece:
> > Let's take into account that the estimates report that comes back from the priority method only speak about priority. These cannot be shown as fees without first multiplying them by the transaction weight.
> 
> Agree that it would be nice to show the final cost, but sheikah already supports relative fee ("Fee per weight unit"), which is the same concept as priority.

We agreed that relative vs. absolute will only be shown if the selected option is "custom". Fees shown in the selector component need to be absolute, paired with the estimated time-to-block.

src/store/wallet.js
```diff
@@ -705,33 +712,32 @@ export default {
         session_id: this.state.wallet.sessionId,
         wallet_id: this.state.wallet.walletId,
         label,
-        fee: standardizeWitUnits(parameters.fee, WIT_UNIT.NANO),
+        fee: `${parseInt(standardizeWitUnits(parameters.fee, WIT_UNIT.NANO))}`,
```
tmpolaczyk, line 715: This converts the priority to an integer, so we lose precision, and any priority lower than 1 will be truncated to 0.

This conversion is needed because the wallet `CreateVttRequest` and similar APIs only accept integer fees (fee has type `u64`). If we knew the weight in advance, we could pass an absolute fee instead of a relative fee, avoiding this problem. Also maybe we could change the wallet API to support fractional priorities. What do you think @aesedepece?

gabaldon: The result of each tier fee shown when selecting the miner fee is the priority already multiplied by the tx weight. When testing local, if the priority < 0, the miner fee sent to the wallet to generate the weighted value is 0 as it does not support < 0 nanowit.

aesedepece: > This converts the priority to an integer, so we lose precision, and any priority lower than 1 will be truncated to 0.

Yesterday I noticed this issue but couldn't locate what was the line actually performing the troublesome truncation.

> This conversion is needed because the wallet `CreateVttRequest` and similar APIs only accept integer fees (fee has type `u64`). If we knew the weight in advance, we could pass an absolute fee instead of a relative fee, avoiding this problem. Also maybe we could change the wallet API to support fractional priorities. What do you think @aesedepece?

Totally for letting the client know the weight, let them do the `weight` by `priority` multiplicarion on their side, and expect them to provide the absolute fee.

aesedepece: > The result of each tier fee shown when selecting the miner fee is the priority already multiplied by the tx weight. When testing local, if the priority < 1, the miner fee sent to the wallet to generate the weighted value is 0 as it does not support < 1 nanowit.

Where did you get the weight from? :thinking: 

With the latest commits on https://github.com/witnet/witnet-rust/pull/2261, the transaction weight is now returned upon calls to `createVtt` and `createDataReq`. Also, fees are now also accepted as string-formatted floats, as priority (aka relative fees) can be less than `1` for reasons.

src/components/SetFee.vue
```diff
@@ -0,0 +1,389 @@
...
+        let transaction
+        if (this.drValues && this.formatedFeeEstimationReport[trait]) {
+          transaction = await this.createDataRequest({
+            parameters: {
+              ...this.drValues,
+              fee: this.formatedFeeEstimationReport[trait].priority,
+              feeType: this.feeType,
+            },
+            request: this.drValues.template.radRequest,
+          })
```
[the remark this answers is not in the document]
aesedepece: You are totally right, my bad!

src/store/wallet.js
tmpolaczyk: > Where did you get the weight from? thinking

The way it works is that it creates a transaction using the relative fee, and the output of that command has the absolute fee.

aesedepece (changes requested):
My two cents from the UX POV:
- [ ] When going back from the fee selection view to the view in which we enter the address and amount, I'm expecting the button to read "Back", not "Cancel" because "Cancel" sounds to me like I want to totally quit creating a transaction.
- [ ] When going back from the fee selection view to the view in which we enter the address and amount, I'm expecting the fields to be completed with the information that I already entered, but right now they go empty and I need to fill them again.
- [ ] We can take this chance to give a helpful caption, placeholder or tooltip to the timelock field, something in the lines of "The recipient will not be able to spend the coins before this date and time".
- [ ] In the fee selector, it's probably a good idea to pre-select "Medium" by default instead of stinky. That will make things better for the users who won't read, won't understand or will simply click "Continue" anxiously. We can also tag it with a "(recommended)" label for assurance.
- #1802. This is not an informative modal but a very interactive one with multiple steps. Not cool if I accidentally close it by clicking on the shaded background, because the data I entered will be lost silently. I think we should only be able to escape from the transaction creation flow through a `x` sign in the corner of the modal.

aesedepece:
>     * [x]  When going back from the fee selection view to the view in which we enter the address and amount, I'm expecting the button to read "Back", not "Cancel" because "Cancel" sounds to me like I want to totally quit creating a transaction.
> 
>     * [ ]  When going back from the fee selection view to the view in which we enter the address and amount, I'm expecting the fields to be completed with the information that I already entered, but right now they go empty and I need to fill them again.

These two points also apply when going from the final confirmation step back to the fee choosing view.

aesedepece:
So nice!

image

aesedepece:
Example with absolute fee (nanowits amount as integer):
```js
{..., "fee":{"Absolute":123}}
```
With relative fee (priority value as float):
```js
{..., "fee":{"Relative":123.456}}
```

aesedepece:
@gabaldon This is working fantastic for me! Only little adjustment I'd made is for the fee selection view to remember your selection if you move forward to the confirmation view, and back to fee selection (it's changing to medium instead of preserving your selection) :eyes:

src/utils.js
```diff
@@ -41,6 +47,19 @@ export function isGrtMaxNumber(value, unit) {
   )
 }
 
+// calculate formated time per block
```
Tommytrg, line 50:
```suggestion
// calculate formated time per interval
```

src/utils.js
```diff
@@ -155,22 +174,26 @@ export function standardizeWitUnits(
...
-  if (amount && amount !== '0') {
-    const num = new BigNumber(amount)
-    const exponent = witUnitConversor[inputUnit][outputUnit]
-    const result = num.times(new BigNumber(10).pow(exponent), 0)
-    if (truncate === -1) {
-      return result.toFixed()
-    } else if (result.cmp(1) === -1) {
-      // result < 1
-      return result.toFixed()
+  try {
```
Tommytrg, line 177: What is throwing an error here?

src/utils.js
```diff
@@ -286,6 +309,14 @@ export function changeDateFormat(timestamp) {
   return `${month} ${day}, ${year} @ ${time}`
 }
 
+//convert to snake case
+export function convertToSnakeCase(value) {
```
Tommytrg, line 313: I'd suggest to use an external library to do this: https://www.npmjs.com/package/snake-case. Is there is a reason to not use it...
 
which is the expected format of `value`? Can contain spaces,dashes or any other separator? Can we add a condition to check if the format is valid?

Tommytrg (changes requested):

aesedepece:
> [...] Only little adjustment I'd made is for the fee selection view to remember your selection if you move forward to the confirmation view, and back to fee selection (it's changing to medium instead of preserving your selection) eyes

This is working perfectly now. Thanks @gabaldon !

src/utils.js
tmpolaczyk: I believe this function is dead code, I don't see it being used anywhere, so let's remove it?

src/utils.js
gabaldon: I can fail when the value passed to new BigNumber(value) is not a number

tmpolaczyk (approved):

src/store/wallet.js
```diff
@@ -391,13 +411,13 @@ export default {
...
-        name !== 'xprv' ||
-        name !== 'seed' ||
-        name !== 'nodeSync'
+        ['uploadFile', 'mnemonics', 'xprv', 'seed', 'nodeSync'].includes(
+          name,
+        ) ||
+        error === 'Validation Error'
       ) {
+        return
+      } else {
```
Tommytrg, line 420: We can remove this else and add the code just below the if

Tommytrg (approved):

PR #1803, opened 2022-09-07
feat(chore): update witnet public nodes

### LINKED ISSUE #1800: Update node urls

gabaldon:
close #1800

### DIFF
--- a/src/background/utils/overwriteWitnetNodeConfiguration.ts
+++ b/src/background/utils/overwriteWitnetNodeConfiguration.ts
@@ -11,21 +11,22 @@ export function overwriteWitnetNodeConfiguration({
   witnetConfigFileName: string
   publicNodeUrls: Array<string>
 }) {
+  const replacement = `node_url = ${JSON.stringify(publicNodeUrls)}\n`
+    .replace("'", '')
+    .trim()
+  const nodeUrlUntilCharacter = (character: string) =>
+    new RegExp('node_url =([^;]*)' + character)
   try {
     fs.writeFileSync(
       path.join(sheikahPath, witnetConfigFileName),
-
       fs
         .readFileSync(path.join(sheikahPath, witnetConfigFileName))
         .toString()
-        .replace(
-          'node_url = "127.0.0.1:21338"',
-          `node_url = ${JSON.stringify(publicNodeUrls)}\n`.replace("'", ''),
-        )
-        .replace('52.166.178.145:21338', `20.126.70.77:21338`),
+        .replace(nodeUrlUntilCharacter('"'), replacement)
+        .replace(nodeUrlUntilCharacter(']'), replacement),
     )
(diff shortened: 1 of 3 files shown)


src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -14,17 +14,20 @@ export function overwriteWitnetNodeConfiguration({
...
         .replace(
-          'node_url = "127.0.0.1:21338"',
+          /node_url =([^;]*)"/,
           `node_url = ${JSON.stringify(publicNodeUrls)}\n`.replace("'", ''),
         )
-        .replace('52.166.178.145:21338', `20.126.70.77:21338`),
+        .replace(
+          /node_url =([^;]*)]/,
+          `node_url = ${JSON.stringify(publicNodeUrls)}\n`.replace("'", ''),
+        ),
```
tmpolaczyk, line 27: Isn't this exactly the same as the previous replace?

```suggestion
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -14,17 +14,20 @@ export function overwriteWitnetNodeConfiguration({
...
     )
   } catch (error) {
+    console.log('error!!!', error)
```
Tommytrg, line 30:
```suggestion
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
gabaldon: It is not, the first one replace the node urls within `""` and the second one within `[]`

tests/unit/src/background/utils/overwriteWitnetNodeConfiguration.spec.ts
```diff
@@ -47,7 +47,7 @@ describe('overwriteConfigFile', () => {
     )
     expect(fs.writeFileSync).toBeCalledWith(
       'sheikah_path/witnet_config_file_name',
-      'node_url = ["20.126.70.77:21338","public_node_url2"]',
+      'node_url = ["public_node_url1","public_node_url2"]\n',
```
tmpolaczyk, line 50: This test is called `it('Should overwrite the outdated node urls')`, so I believe it was to ensure that only public nodes have their address changed, because users are allowed to set other nodes in this field, and in that case we don't want to delete their node addresses.

gabaldon: Sheikah overwrites all node addresses every time there is a new Sheikah update, just in case the version is not compatible. Anyway, this test description is not accurate. I will change it to `Should overwrite a configuration with several nodes according to the parameters given`

aesedepece (approved):

PR #1816, opened 2022-10-24
fix(background): enable electron devtools

Tommytrg:
Open a new window for the electron devtools. This workaround is necessary due to an issue in an old version of electron: https://github.com/electron/electron/issues/20069#issuecomment-586642795

### DIFF
--- a/src/background/appManager.ts
+++ b/src/background/appManager.ts
@@ -9,6 +9,8 @@ declare const __static: string
 export class AppManager {
   public status: Status
   public win: BrowserWindow | null = null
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
This works perfectly, but it is only possible to open the devtools at startup. The `ctrl-shift-k` shortcut does not work. So ideally, this PR should fix that shortcut, allow opening the devtools manually, and disable the automatic open at startup.

Tommytrg:
This is working with the current version

PR #1826, opened 2022-11-08
refactor(ui): avoid store class instances in vuex

Tommytrg:


gabaldon (changes requested):
You should also update context.rootState.wallet.api.getItem  and context.rootState.wallet.api.saveItem in rad.js. The wallet api is not anymore in wallet.js

PR #1845, opened 2022-12-20
Fix overwrite

gabaldon:


src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -1,30 +1,60 @@
...
   sheikahPath,
   witnetConfigFileName,
   publicNodeUrls,
+  oldWitnetNodes,
 }: {
   sheikahPath: string
   witnetConfigFileName: string
   publicNodeUrls: Array<string>
+  oldWitnetNodes: Array<string>
 }) {
```
Tommytrg, line 30: We are passing always the same arguments to this function. I like being able to pass them to be able to test the function easily, but could we add a default argument value?

```suggestion
const defaultNodeConfiguration = {
    sheikahPath: SHEIKAH_PATH,
    witnetConfigFileName: WITNET_CONFIG_FILE_NAME,
    publicNodeUrls: URLS_PUBLIC_WITNET_NODES,
    oldWitnetNodes: OLD_WITNET_NODE_IPS,
}

export function overwriteWitnetNodeConfiguration({
  sheikahPath,
  witnetConfigFileName,
  publicNodeUrls,
  oldWitnetNodes,
}: {
  sheikahPath: string
  witnetConfigFileName: string
  publicNodeUrls: Array<string>
  oldWitnetNodes: Array<string>
} = defaultNodeConfiguration) {
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -1,30 +1,60 @@
...
+
+export function hasOldWitnetIp({
+  oldWitnetNodes,
+  currentNodeConfiguration,
+}: {
+  oldWitnetNodes: Array<string>
+  currentNodeConfiguration: Array<string>
+}) {
+  return currentNodeConfiguration.some(ip => oldWitnetNodes.includes(ip))
+}
```
Tommytrg, line 17: This function doesn't need to know about our business  logic
```suggestion
export function existIntersection(
  arr1: Array<string>,
  arr2: Array<string>,
)) {
  return arr1.some(item => arr2.includes(item))
}
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -1,30 +1,60 @@
...
+  oldWitnetNodes,
 }: {
   sheikahPath: string
   witnetConfigFileName: string
   publicNodeUrls: Array<string>
+  oldWitnetNodes: Array<string>
 }) {
   const replacement = `node_url = ${JSON.stringify(publicNodeUrls)}\n`
     .replace("'", '')
     .trim()
```
tmpolaczyk, line 33: No need to calculate this string if hasOldWitnetIp returns false, so let's move it inside the if.

tests/unit/src/background/utils/overwriteWitnetNodeConfiguration.spec.ts
```diff
@@ -28,20 +29,17 @@ describe('overwriteConfigFile', () => {
     )
   })
 
-  it('Should overwrite the configuration with several nodes according to the parameters given', () => {
+  it('Should overwrite config only if there is only one old witnet node ip', () => {
```
tmpolaczyk, line 32:
```suggestion
  it('Should overwrite config if there is only one old witnet node ip', () => {
```

tmpolaczyk (approved):

Tommytrg (approved):

PR #1847, opened 2022-12-21
force overwrite when first time using sheikah

gabaldon:


src/background/walletManager.ts
```diff
@@ -245,6 +242,10 @@ export class WalletManager {
   public async runWallet() {
     await sleep(3000)
     console.info('Running wallet...')
+    // Is first time running Sheikah
+    if (!this.existDirectory) {
```
Tommytrg, line 246:
```suggestion
    if (!this.existDirectory) {
      // Is first time running Sheikah
```

src/background/walletManager.ts
```diff
@@ -87,6 +87,7 @@ export class WalletManager {
         console.error('An error occured trying to read version file', err)
       }
     }
+
```
Tommytrg, line 90:
```suggestion
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -62,6 +60,17 @@ export function overwriteWitnetNodeConfiguration(
...
+      overwriteConfig()
+    } else if (
+      configNodeIps &&
+      existIntersection({
+        oldWitnetNodes: oldWitnetNodes,
+        currentNodeConfiguration: configNodeIps,
+      })
+    ) {
+      overwriteConfig()
+    }
```
Tommytrg, line 73:
```suggestion
if (
      force || (configNodeIps &&
      existIntersection({
        oldWitnetNodes: oldWitnetNodes,
        currentNodeConfiguration: configNodeIps,
      }))
    ) {
      overwriteConfig()
    }
```

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -62,6 +60,17 @@ export function overwriteWitnetNodeConfiguration(
         configFile.replace(nodeUrlConfigRegex, replacement),
       )
     }
+    if (force) {
```
Tommytrg, line 63: We only need an `if` condition here

src/background/utils/overwriteWitnetNodeConfiguration.ts
```diff
@@ -45,15 +48,10 @@ export function overwriteWitnetNodeConfiguration(
...
       .replaceAll('"', '')
       .split(',')
-    if (
-      configNodeIps &&
-      existIntersection({
-        oldWitnetNodes: oldWitnetNodes,
-        currentNodeConfiguration: configNodeIps,
-      })
-    ) {
+    const overwriteConfig = () => {
```
Tommytrg, line 54: We don't need to extract this logic into a function. We can write it directly inside the `if`

Tommytrg (approved):

PR #1858, opened 2023-05-22
Allow to import in myWitWallet from XPRV using a QR code

Tommytrg:


src/components/ExportXprvModal.vue
```diff
@@ -0,0 +1,145 @@
...
+    :title="$t('close_session_info_title')"
+    width="50%"
+    :visible="true"
+    :show-close="true"
+    :close-on-click-modal="false"
+    @close="close"
+  >
+    <div slot="title" class="title-container">
+      <font-awesome-icon class="icon" icon="info-circle" />
+      <p class="title">{{ $t('export_xprv_title') }}</p>
```
gabaldon, line 13: You need to add the copy to locales files

gabaldon (changes requested):

src/components/ExportXprvModal.vue
Tommytrg: `export_xprv_title` was wrong. Te correct key is `export_xprv_modal_title`

gabaldon (approved):

PR #1865, opened 2023-07-12
fix: prevent internal keychain from skipping indexes

### LINKED ISSUE #1864: Generate transactions without persisting change addresses
The bug was reported and fixed in  [witnet-rust #2385](https://github.com/witnet/witnet-rust/pull/2385), but to make it work the `preview: true` flag should be passed to `createVTT` and `createDRT`.

aesedepece:
Fix #1864 
Rel https://github.com/witnet/witnet-rust/pull/2385

### DIFF
--- a/src/components/SendTransaction/SetFee.vue
+++ b/src/components/SendTransaction/SetFee.vue
@@ -232,6 +232,7 @@ export default {
             ...this.vttValues,
             fee: this.formatedFeeEstimationReport[tier].priority,
             feeType: this.feeType,
+            preview: true,
           })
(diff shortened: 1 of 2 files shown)


Tommytrg:
We had a problem with the minimum `witnet-rust` version allowed in Sheikah due to the fee estimation feature and we fixed it by increasing the version. In addition, there was a problem related to the `witnet-rust` binary using 1.6 or higher but I can't reproduce it. It could be only related to the windows binary.

aesedepece:
This is confirmed to work fine with the latest changes in https://github.com/witnet/witnet-rust/pull/2385

Tommytrg (approved):

PR #1876, opened 2024-02-16
ci: fix create draft release

gabaldon:


package-lock.json
```diff
@@ -0,0 +1,12615 @@
+{
```
Tommytrg: Could we delete this file?

Tommytrg (approved):

